PR #12, opened 2017-07-17
make end_basic_block public

MaikKlein:
I assume this should be public?

### DIFF
--- a/rspirv/mr/builder.rs
+++ b/rspirv/mr/builder.rs
@@ -212,7 +212,7 @@ impl Builder {
         Ok(id)
     }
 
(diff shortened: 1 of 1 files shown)


MaikKlein:
I signed it!

antiagainst:
Thanks for the contribution, @MaikKlein!

Actually `end_basic_block` should only be used when creating terminator instructions like branch, conditional branch, switch etc., e.g., https://github.com/google/rspirv/blob/master/rspirv/mr/build_terminator.rs#L23.

Therefore, it should not be called by outside. :)

antiagainst:
I'll close this. :)

PR #13, opened 2017-07-17
Add assemble_bytes

MaikKlein:
Adds

```
pub trait Assemble {
    /// Assembles the current object and returns the binary code.
    fn assemble(&self) -> Vec<u32>;
    fn assemble_bytes(&self) -> Vec<u8> {
        use std::mem;
        self.assemble()
            .iter()
            .flat_map(|val| {
                (0..mem::size_of::<u32>()).map(move |i| ((val >> (8 * i)) & 0xff) as u8)
            })
            .collect()
    }
}
```

I think it would be useful to directly output the byte code.

### DIFF
--- a/rspirv/binary/assemble.rs
+++ b/rspirv/binary/assemble.rs
@@ -22,8 +22,7 @@ pub trait Assemble {
     fn assemble(&self) -> Vec<u32>;
     fn assemble_bytes(&self) -> Vec<u8> {
         use std::mem;
-        b.module()
-            .assemble()
+        self.assemble()
             .iter()
             .flat_map(|val| {
                 (0..mem::size_of::<u32>()).map(move |i| ((val >> (8 * i)) & 0xff) as u8)
--- a/rspirv/binary/assemble.rs
+++ b/rspirv/binary/assemble.rs
@@ -21,6 +21,7 @@ pub trait Assemble {
     /// Assembles the current object and returns the binary code.
     fn assemble(&self) -> Vec<u32>;
     fn assemble_bytes(&self) -> Vec<u8> {
+        use std::mem;
         b.module()
             .assemble()
             .iter()
--- a/rspirv/binary/assemble.rs
+++ b/rspirv/binary/assemble.rs
@@ -20,6 +20,15 @@ use utils::num::{bytes_to_u32_le, f32_to_u32};
 pub trait Assemble {
     /// Assembles the current object and returns the binary code.
     fn assemble(&self) -> Vec<u32>;
+    fn assemble_bytes(&self) -> Vec<u8> {
+        b.module()
+            .assemble()
+            .iter()
+            .flat_map(|val| {
+                (0..mem::size_of::<u32>()).map(move |i| ((val >> (8 * i)) & 0xff) as u8)
+            })
+            .collect()
+    }
 }
 
 impl Assemble for mr::ModuleHeader {


MaikKlein:
@antiagainst Sure, but I have one question. Should modules be comparable? I would like to serialize a module and parse it again and then compare the two modules for equality.

MaikKlein:
Actually, it is also not correct. I can serialize it and parse it again with `parse_bytes` and it prints the same module.

But I can not feed the generated spirv into `spirv-cross`.

``` 
pub fn resave_spirv<T: AsRef<Path>>(t: T) {
    use std::io::Read;
    use rspirv::binary::{Assemble};
    use std::io::Write;
    use std::mem;
    let path = t.as_ref();
    let module = {
        let mut file = File::open(path).unwrap();
        let mut v = Vec::new();
        file.read_to_end(&mut v);
        let mut loader = rspirv::mr::Loader::new();
        rspirv::binary::parse_bytes(&v, &mut loader);
        loader.module()
    };
    println!("{:?}", module);
    let bytes: Vec<u8> = module
        .assemble()
        .iter()
        .flat_map(|val| {
            (0..mem::size_of::<u32>()).map(move |i| ((val >> (8 * i)) & 0xff) as u8)
        })
        .collect();
    let mut file = File::create(path).unwrap();
    file.write_all(&bytes);
}
```

So I assume that I am writing it to the wrong format.

Although the spec says

>Tip
Endianness: A module is defined as a stream of words, not a stream of bytes. However, if stored as a stream of bytes
(e.g., in a file), the magic number can be used to deduce what endianness to apply to convert the byte stream back to a
word stream.

This is how the spirv file looks, generated from glslang. https://gist.github.com/MaikKlein/a160cef6894235155dd85ac17ef023ec

And this is the spirv file after I call `resave_spirv` https://gist.github.com/MaikKlein/66db7e1a3479039dc2aa1f1d0aee66bd

Which is completely different.

antiagainst:
I'm curious why that happens. It seems fine on my machine?

I did some experiment by applying the following patch:

```diff
diff --git a/dis/main.rs b/dis/main.rs
index 1f5b833..faed959 100644
--- a/dis/main.rs
+++ b/dis/main.rs
@@ -17,8 +17,10 @@ extern crate rspirv;
 
 use std::fs;
 use std::io::Read;
+use std::io::Write;
+use std::mem;
 
-use rspirv::binary::Disassemble;
+use rspirv::binary::Assemble;
 
 fn main() {
     let matches = clap::App::new("rspirv-dis")
@@ -34,7 +36,18 @@ fn main() {
     f.read_to_end(&mut buffer).expect("cannot read file");
 
     match rspirv::mr::load_bytes(&buffer) {
-        Ok(module) => println!("{}", module.disassemble()),
+        Ok(module) => {
+            let bytes: Vec<u8> = module
+                .assemble()
+                .iter()
+                .flat_map(|val| {
+                    (0..mem::size_of::<u32>()).map(move |i| ((val >> (8 * i)) & 0xff) as u8)
+                })
+                .collect();
+
+            let mut file = fs::File::create("input.resave").unwrap();
+            file.write_all(&bytes).unwrap();
+        },
         Err(err) => println!("{}", err),
     }
 }
```
Then using `rspirv-dis` to re-assemble a SPIR-V binary. The contents in `input.resave` are the same as input, except for headers.

MaikKlein:
Yes I have the same behavior. The problem here is just that spirv-cross can't seem to read the outputs.

```

➜  examples git:(master) ✗ cat test.vert
#version 400

// int test(int a, int b){
//   return a + b;
// }

void main() {
  //int i = test(10, 5);
  uint a = 1;
}
➜  examples git:(master) ✗ glslangValidator -V test.vert
test.vert
Warning, version 400 is not yet complete; most version-specific features are present, but some are missing.

➜  examples git:(master) ✗ spirv-cross vert.spv
#version 400
#ifdef GL_ARB_shading_language_420pack
#extension GL_ARB_shading_language_420pack : require
#endif

void main()
{
    uint a = 1u;
}
```

But if I then load vert.spv and save it to `input.resave`

```
➜  examples git:(master) ✗ spirv-cross input.resave
terminate called after throwing an instance of 'spirv_cross::CompilerError'
  what():  No function currently in scope
[1]    5553 abort (core dumped)  spirv-cross input.resave
```

Does it work for you? I am not sure what the problem here is. I need to investigate a bit further.

MaikKlein:
Actually I found the problem. The only reason why this not works is the following line

```
uint i = 4; // Same goes for float and int
``` 

Every other shader that I have tested works. 

```
#version 400

uint i = 4;
void main()
{
}
```

This works. So I guess there might be a problem with function local variables?

antiagainst:
Ouch, it's a bug in the loader definitely.

`OpVariable` can happen both in the module and in a function. But in the current loader implementation, [all are put in the module's variable section](https://github.com/google/rspirv/blob/master/rspirv/mr/loader.rs#L147). That is apparently wrong.

MaikKlein:
I have disassembled both spv

https://gist.github.com/MaikKlein/8b0035a8fbf1ca01b2e3d998a737da12#file-gistfile1-txt-L53

and 

https://gist.github.com/MaikKlein/d01b323584dbe38cec8f3acba39f27bd#file-gistfile1-txt-L51

I think the problem is that 

```
%i = OpVariable %_ptr_Function_uint Function
```
is declared before the function `main`.

Sorry for side stepping this PR

antiagainst:
Thanks very much for spotting this, @MaikKlein!

Would you like to fix this or actually you'd like me to fix it? It should be relatively easy, just need to make sure that `self.function` is `None` before `self.module.types_global_values.push(inst)`.

MaikKlein:
You should better do that yourself, I am not yet familiar with your code base.

antiagainst:
Sure. I'll fix it soon.

antiagainst:
Should be fixed by https://github.com/google/rspirv/commit/301e6a57e1ad723a7185e8f86312fdd8270922bc

kvark:
@MaikKlein can this be revived?

kvark:
Closing due to inactivity.
Also, gfx-rs and wgpu-rs now work with `[u32]` fwiw.

PR #26, opened 2017-10-12
Update to bitflags 1.0.

waywardmonkeys:
This uses associated constants, so it requires Rust 1.20 or later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 name = "spirv_headers"
 version = "1.1.6"
 dependencies = [
- "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "bitflags 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "num 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-derive 0.1.41 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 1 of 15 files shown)


waywardmonkeys:
This probably requires a full version number bump on the `spirv-headers` crate as the exported values are no longer the same.  Hopefully, `bitflags` 1.0 is the future and this would need to happen anyway. This should also result in better generated documentation as the constants will be grouped together appropriately.

waywardmonkeys:
`clap` has an open pull request to update it to `bitflags` 1.0 as well: https://github.com/kbknapp/clap-rs/pull/1063

antiagainst:
Looks good to me. Thanks a lot, @waywardmonkeys!

Yes, I agree that a version bump is needed for this change. Currently the version of `spirv_headers` is just the version of SPIR-V it encodes. So that means upgrading to use SPIR-V 1.2. (Actually I still kinda want to publish one version of `spirv_headers` for SPIR-V 1.0 headers at least in case somebody need it. Vulkan 1.0 requires SPIR-V 1.0.)

The nasty thing is that now we have cases that two capabilities are aliases to each other and they map to the same discriminator, e.g., [both StorageBuffer16BitAccess and StorageUniformBufferBlock16 maps to 4433](https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/1.2/spirv.core.grammar.json#L5652), which requires some change to the codegen phase. But associated constants seems a nice solution to that.

I'd like to push two more versions for 1.1 to catch up with the spec version, and then migrate to 1.2. So may defer merging this pull request to after that. :)

antiagainst:
Rebased and merged as https://github.com/google/rspirv/commit/0ebff272db9dfa9710eea6fbd78e0f7977980375

PR #29, opened 2018-04-20
Strip the prefix from the Dim enum on printing

leops:
Tools like spirv-as fail on the invalid enum in the generated assembly.

I just changed the Display implementation of Operand because the change had the smallest footprint, but maybe a better fix would be manually implementing Debug for the Dim enum ?


antiagainst:
Thanks @leops! Yep, this is reasonable. But we also need to update the file in `codegen/` that generates this file. I've taken your commit and modified it and pushed as https://github.com/google/rspirv/commit/9af5b149fc438db1d70549d8c9bb19831197d6d4.

PR #31, opened 2018-08-20
Fix disassembly version

grovesNL:
The default version has changed from 1.0 to 1.3, so the `assert_eq` in the README fails.


grovesNL:
Actually this isn't the issue, so don't merge this. It seems like the version in the header is skipped during `parse_header`. I'll look into it.

grovesNL:
I've updated `parse_header` to parse the version and reused the same `Word` -> `(u8 u8)` function which already existed (just moved it to a separate module). This fixes the README and some tests/doctests which seemed incorrect previously.

Let me know what you think.

PR #36, opened 2019-06-06
Version bump `num` to 0.2 for spirv_headers

icefoxen:
Did my best to leave all other dependencies the same version.  Not sure if this counts as an API break for you, or if you even want it.  I was just building gfx-hal and wondering "what in the world is using `rustc-serialize` anyway?", and eventually traced it to num, through this.


antiagainst:
Awesome, thanks!

PR #37, opened 2019-07-04
Enable Windows on Travis CI

kvark:
... and remove appveyor


antiagainst:
Based on my prior experience, Appveyor's service is far reliable than Travis. So I'm a bit skeptical of fully using Travis. What's your experience here? 

Besides, on the Appveyor side, I've paid service so it can start two jobs concurrently.

kvark:
The only incentive here was to reduce the number of services we rely on. I.e. with both Travis and AppVeyor involved, the blockout of any of these would cause PR merging to stall.

Since gfx-rs switched to Travis on Windows, we did have a few sporadic cases of network resources being unavailable from the script, but nothing serious. In terms of performance (re: paid AppVeyor), Windows targets are not a blocker. OSX often is, since the servers are more busy and the builds start much later than we request (despite the total build length being smaller than on windows, see https://travis-ci.org/gfx-rs/gfx/builds/552451138).

So that's my reasoning. I'm fine with whatever call you make on this PR.

PR #39, opened 2019-07-15
CI and Reame updates

kvark:
This is a successor to #37 
It removes AppVeyor, adds bors and rustfmt configurations, and updates the README


kvark:
@antiagainst oh, I somehow didn't expect Google CLA bot to show up now. Do we need to explicitly disable it somehow?

kvark:
@antiagainst nvm, I think I cleaned up the settings now (let me know if I screwed up something important!). Going to delete the bot comment from here.

.travis.yml
```diff
@@ -1,37 +1,25 @@
...
+      rust: nightly
+    - os: osx
+      rust: stable
+    - os: windows
+      rust: stable
 
 branches:
-  only:
-    - master
+  except:
```
antiagainst: Why this change? Do we have some conventions for using branches in gfx-rs?

kvark:
Re: branches ci.

Unless it's a personal repository, all the branches are expected to be long-lived. Any short feature branch would be in a fork. The only branches we expect in upstream are for releases (so that we can backport patches) and long-lived features (like when gfx-hal was developed). All of these receive prs and need to be Ci-checked, just like "master".

The excluded staging branch is for bors, which creates it as an implementation detail.

antiagainst:
I'm not sure that all the branches must be long-lived. There are situations that we'll need to communicate hot fixes for some bug to multiple developers and those should be only useful for a short time. I think we should not make the guarantee that we need to always maintain them.

But anyway, these are different concerns. Whatever our policy is here, it does not block us from testing it. :) I was basically curious so asked.

kvark:
> I'm not sure that all the branches must be long-lived. There are situations that we'll need to communicate hot fixes for some bug to multiple developers and those should be only useful for a short time. I think we should not make the guarantee that we need to always maintain them.

If, in some case, you want to do a quick web-based fix (that uses a local branch in upstream), it's not the end of the world to have it CI-checked. As long as it doesn't happen too often and doesn't strain CI visibly, that is :)

PR #40, opened 2019-07-15
Rename all generated sources to start with gen_

kvark:
I think it's helpful for developers to be able to recognize instantly if a file is auto-generated (and so can't be modified by hands) or not. Enforcing a naming scheme is a way to achieve that. I'm open to other ideas :)


antiagainst:
SGTM to have a prefix in the filename to distinguish. One nit, can we use `autogen_` instead? I believe it's more clear that way.

kvark:
Sure, thanks for reviewing!
I'll change to "autogen" now. FYI, I also considered "generated_" since this is what Gecko uses with cbindgen when gluing with WebRender.

kvark:
bors r=antiagainst

PR #44, opened 2019-07-16
Refactor the Parser

kvark:


kvark:
@antiagainst you can see in this PR that I pushed every commit separately, and there is a CI run on each. So if/when it merges, we know that every commit will be good.

antiagainst:
@kvark: yes I see. Thanks for the due diligence! I'll rebase and merge this PR to keep these two commits separate then. :)

Issue #5, opened 2017-02-11
Structured representation


antiagainst:
Currently there is only a plain data representation (DR) implemented in this project. While it  introduces little overheads when working with the binary parser and is straightforward for investigating the contents of SPIR-V binaries, it has quite limited functionalities. Instructions are not interconnected; types are not hierarchicalized; metadata (decorations, debug instructions, etc.) are not linked to the targets; etc. It is inconvenient to build future tasks like validation, optimization, and translating from front end languages on top of this DR. So a structured representation (SR) should be introduced, containing:

* A type hierarchy
* A constant hierarchy
* Common language constructs: Module, Function, BasicBlock, Instruction, etc.
* Representing all SPIR-V instructions
* Decoration struct. This will be attached to types, instructions, etc.

### Type hierarchy

We can either implement it using 1) enums or 2) marker traits with multiple implementing structs.

1. Types are frequently used, it would be nice to reduce the overheads. This favors enums since marker traits leads to trait objects.
2. It reads better and is more easy to use (without excessive `match`s). Also easy to attach data and methods (for like, getting constants from a type, composing types, etc.) on each type.

I'm leaning towards 2. Let's worry about performance later.

```rust
// sketch

trait Type {
  is_numerical(&self) -> bool;
  is_aggregate(&self) -> bool;
  is_composite(&self) -> bool;
  ...
}
trait ScalarType { ... }
trait AggregateType { ... }

struct FloatType {
  bitwidth: u32,
}
struct VectorType {
  element_type: &ScalarType,
  size: u32,
}
struct ArraryType { ... }

impl ScalarType for IntType { ... }
impl AggreateType for VectorType { ... }
```

### Representing SPIR-V instructions

A single flexible instruction struct that can represent all SPIR-V instructions is inferior to a dedicated instruction struct for each SPIR-V instruction: it loses the structure naturally encoded in dedicated instruction struct and will need validation for operand layout correctness, etc. A huge enum containing all possible instructions is ugly.  It seems there is no way to go around trait objects, which is avoided by the Rust community to the best. But it's the natural way of fulfilling this kind of task.

```rust
// Sketch
 
trait Instruction {
  result_id(&self) -> Option<u32>;
  result_type(&self) -> Option<&Type>;
  ...
}

struct IAddInst {
  result_type: &Type,
  result_id: u32,
  op1: &Instruction,
  op2: &Instruction
}

impl Instruction for IAddInst { ... }
```
[...]

PR #46, opened 2019-07-16
[WIP] Structured items

kvark:
The PR contains core missing pieces of #5: rich instructions and top-level items, such as `Module`.


kvark:
I think the biggest thing missing is basically a `function(Context, DataInstruction) ->StructuredInstruction`. It could be implemented as a few auto-generated methods:
```rust
impl Context {
  pub fn lift_type(&mut self, raw: &Instruction) -> TypeToken;
  pub fn lift_extension(&mut self, raw: &Instruction) -> Extension;
  pub fn lift_operand(&mut self, <something>) -> Operand;
  ..etc
}
```

kvark:
Looking at this some more... I'm able to generate the lift() functions that process and match operands, aka:
```rust
    pub fn lift_memory_model(
        &mut self,
        raw: &mr::Instruction,
    ) -> Result<structs::MemoryModel, LiftError> {
        if raw.class.opcode as u32 != 14u32 {
            return Err(LiftError::OpCode);
        }
        let mut operands = raw.operands.iter();
        Ok(structs::MemoryModel {
            addressing_model: (match operands.next() {
                Some(&mr::Operand::AddressingModel(ref value)) => Some(value.clone()),
                Some(_) => Err(OperandError::Wrong)?,
                None => None,
            })
            .ok_or(OperandError::Missing)?,
            memory_model: (match operands.next() {
                Some(&mr::Operand::MemoryModel(ref value)) => Some(value.clone()),
                Some(_) => Err(OperandError::Wrong)?,
                None => None,
            })
            .ok_or(OperandError::Missing)?,
        })
    }
```

My problem is that we need to enforce the semantics that is in the spec but nowhere to be seen in the schema, e.g.:
> Entry Point must be the Result <id> of an OpFunction instruction.
> Interface is a list of <id> of global OpVariable instructions

In the schema they are just `IdRef` things:
```json
        { "kind" : "IdRef",                              "name" : "'Entry Point'" },
        { "kind" : "LiteralString",                      "name" : "'Name'" },
        { "kind" : "IdRef",          "quantifier" : "*", "name" : "'Interface'" }
```

So there appears to be incomplete information in the schema to build the proper structured API. We have 2 ways to proceed from here:
  1. just implement all/most instructions manually, use the exact typing corresponding to the spec
  2. associate operand names with types in the code generator

@antiagainst do you have a feel on which approach would work best for us? ^

kvark:
@antiagainst I'm going with approach (2) currently, i.e.:
```rust
        match operand.name.trim_matches('\'') {
            "Length" => quote! { ConstantToken },
            "Entry Point" => quote! { FunctionToken },
            "Interface" => quote! { VariableToken },
            _ => quote! { TypeToken },
        }
```
So far, it seems to me that naming is fairly consistent and repetitive, so adding such a match is still a big win over manually writing the code.

Another issue I'm facing is that content in SPIRV appears to not be sequential. An OpEntryPoint point to OpFunction, for example. This is annoying because we want to be populating our context graph with function/instruction/constant/etc tokens, and we can use one unless we processed it first. I'll try to see if we can process all the functions first, before going for the mode setting ops, and whether this is big issue within a function/block as well.

antiagainst:
Yes the JSON grammar was not a full-fledged encoding of the syntax validation rules. (My bad initially defining it that way; it was more meaning to be supporting a data driven approach in SPIRV-Tools.) And you pointed out the most noticeable one: IdRef. I initially thought about adding more sub cases to it but never got to it.

Generally I'm in favor of more autogen. (You can know that which all the codegen stuff I've in this project; although codegen means another thing in compiler so we should probably rename it. ;-P) So matching names as a start sounds good to me. I think a more principled solution would be to enhance the SPIR-V JSON grammar to encode this info. (I'm actually needing other stuff that is not in the upstream; category is added by me for this repo specifically and so you see we are not using an upstream grammar here, which is bad because need to manually merge with each version update.)

antiagainst:
There are a few places that SPIR-V has forward references. The [sections](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_logicallayout_a_logical_layout_of_a_module) before types and global values are all "metadata" instructions; meaning they attach additional information to some other instruction. So frequently they do forward referencing, esp. entry point, execution mode, debugging instruction, annotations. (We can also have `OpTypeForwardPointer` forward references a type...) Sounds good to me to parse the "main" instructions first (the types, global values, and functions) and then getting back to those instructions before. (We have the full MR (which should be named into as DR) representation here so it should be doable.) Also we should fold those metadata instructions into their main instruction if possible. That's why I have a decoration field for most of the instructions.

kvark:
This is obviously incomplete, WIP, etc, but at least it's at the point where it compiles. I don't think having more stuff piled up on top would be a good way to merge later. @antiagainst - do you think you could review this and consider landing if it's done in the preferred way?

codegen/sr.rs
```diff
@@ -21,25 +21,19 @@ use proc_macro2::{Ident, Span, TokenStream};
...
-    } else if kind == "PairLiteralIntegerIdRef" {
-        quote! { (u32, spirv::Word) }
-    } else if kind == "PairIdRefLiteralInteger" {
-        quote! { (spirv::Word, u32) }
-    } else if kind == "PairIdRefIdRef" {
-        quote! { (spirv::Word, spirv::Word) }
-    } else {
-        let kind = Ident::new(kind, Span::call_site());
-        quote! { spirv::#kind }
+    match kind {
```
antiagainst: This is just refining existing implementation and not related to the new feature we are implementing here, right? In the future, it would be nice to have separate PRs.

codegen/sr.rs
```diff
@@ -280,51 +271,192 @@ pub fn gen_sr_type_creation(grammar: &structs::Grammar) -> String {
...
+                    vec.push(value);
+                }
+                vec
+            }
+        },
+        other => panic!("wrong quantifier: {}", other),
+    }
+}
+
+pub fn gen_sr_structs_and_lifts(grammar: &structs::Grammar) -> (String, String) {
```
antiagainst: Can we have some comments regarding this function?

(I wasn't doing good on this previously given it's just me playing with this repo. Now it's intended to be developed by more people so some documentation could really save newcomers efforts in understanding existing codebase. :)

rspirv/Cargo.toml
```diff
@@ -20,6 +20,7 @@ travis-ci = { repository = "gfx-rs/rspirv" }
 num = "0.2"
 derive_more = "0.7"
 clippy = { version = "0.0", optional = true }
+fxhash = "0.2"
```
antiagainst: Nit: can you sort these dependencies alphabetically?

codegen/sr.rs
```diff
@@ -260,14 +254,11 @@ pub fn gen_sr_type_creation(grammar: &structs::Grammar) -> String {
...
             quote! {
-                pub fn #func_name #param_list -> TypeToken {
-                    let t = Type { ty: TypeEnum::#symbol #init_list, decorations: Vec::new() };
-                    if let Some(index) = self.types.iter().position(|x| *x == t) {
-                        TypeToken::new(index)
-                    } else {
-                        self.types.push(t);
-                        TypeToken::new(self.types.len() - 1)
-                    }
+                pub fn #func_name #param_list -> Token<Type> {
```
antiagainst: Here too. It would be nice to have a separate CL for this change and the autogen touched files.

rspirv/sr/context.rs
```diff
@@ -12,11 +12,94 @@
...
+    items::{Function, Variable},
+    structs,
+    LiftError, OperandError,
+};
+use crate::mr;
+
+type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
+type Index = u32;
+
+/// A structure holding some kind of a structured SPIRV data that can be
```
antiagainst: Nit: SPIR-V is the formal spelling. Unless we are in code where we cannot use -, let's use the formal name when possible.

antiagainst: ... some kind of SPIR-V constructs that can ...

rspirv/sr/context.rs
```diff
@@ -29,24 +112,55 @@ use sr::types::{StructMember, TypeEnum};
...
+        &self.types.map[&token.index]
+    }
+
+    /// Returns the reference to the real constant represented by the given token.
+    pub fn get_constant(&self, token: Token<Constant>) -> &Constant {
+        // Note: we assume the vector doesn't shrink so we always have a valid index.
+        &self.constants.map[&token.index]
+    }
+
+    pub fn get_function(&self, token: Token<Function>) -> &Function {
```
antiagainst: Can we have documentations for these public methods?

rspirv/sr/instructions.rs
```diff
@@ -0,0 +1,7 @@
+use spirv;
+use super::{
+	context::Token,
```
antiagainst: Format

rspirv/sr/items.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl From<LiftError> for ConvertionError {
+    fn from(error: LiftError) -> Self {
+        ConvertionError::Lift(error)
+    }
+}
+
+impl Module {
+    pub fn from_data(module: &mr::Module) -> Result<Self, ConvertionError> {
```
antiagainst: Nit: from_data_representation

rspirv/sr/mod.rs
```diff
@@ -14,12 +14,38 @@
...
 mod context;
-mod autogen_decoration;
+mod instructions;
+mod items;
+pub mod structs;
 mod types;
+
+
+#[derive(Clone, Debug)]
+pub enum OperandError {
```
antiagainst: Can we have documentation for public APIs?

rspirv/sr/structs.rs
```diff
@@ -0,0 +1,7 @@
+use spirv;
```
antiagainst: `structs` is not a very descriptive name for this module. It's too general and it implies data representation (which is another representation). Can we have a better one? What about `op`?

rspirv/sr/autogen_instruction.rs
```diff
@@ -25,656 +68,627 @@ enum Instruction {
...
     },
-    Extension {
-        name: String,
-    },
-    ExtInstImport {
-        name: String,
-    },
     ExtInst {
-        set: spirv::Word,
+        set: Token<Type>,
```
antiagainst: Shouldn't all these be `Token<Instruction>`?

kvark:
@antiagainst This PR becomes increasingly difficult to advance without either of #49 or #55 to be merged.

antiagainst:
Just wanted to double check here. So this PR is not expected to be reviewed but expected to be split into smaller ones right?

kvark:
@antiagainst correct. It's technically in the "Draft" status, which isn't ready for review.
I think I'm getting close to catching up to this PR...

antiagainst:
Cool, thanks!

Jasper-Bekkers:
This PR has been stale for about a year, mind if we close it?

Issue #23, opened 2017-08-21
Threaded code generation with mr::Builder


MaikKlein:
I currently want to multi-thread my code generation, but Builder is inherently mutable. The simple solution would be to just wrap it in a mutex, but that would kill the performance.

One solution that I am thinking of is to create a separate builder object for every node that I traverse, then I would return the builder object and merge them together in the correct order.

The only problem that I see are the ids. If I would want to split the builder object into many small builder objects I would need to keep track of the correct IDs.

```
#[derive(Default)]
pub struct Builder {
    module: mr::Module,
    next_id: Arc<AtomicUsize>, // <= from u32 to Arc<AtomicUsize>,
    function: Option<mr::Function>,
    basic_block: Option<mr::BasicBlock>,
}
```
One solution would be to use an `Arc<AtomicUsize>`, this could also made generic.

And and `append` function to merge them together. `builder.append(other_builder);`

What are your thoughts?

antiagainst:
The above sounds good to me. Please feel free to modify and send pull requests! :)

Just keep it as `Arc<AtomicUsize>` would be fine for me. Let's change it to a more generic solution when demanded.

Nit: `append()` sounds more like attaching at the end of the builder; we actually want `merge()`.

PR #49, opened 2019-07-21
sr: Generic Storage and Token linked to SPIR-V

kvark:
Part of #46 that makes our tokens generic, in preparation to having more variations of them.

### DIFF
--- a/rspirv/sr/context.rs
+++ b/rspirv/sr/context.rs
@@ -12,11 +12,101 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::spirv;
+use std::{
+    collections::HashMap,
+    hash::BuildHasherDefault,
+    marker::PhantomData,
+};
 
-use super::{Type, TypeToken, Constant, ConstantToken};
-use crate::sr::constants::ConstantEnum;
-use crate::sr::types::{StructMember, TypeEnum};
+use fxhash::FxHasher;
+
+use crate::{
+    spirv,
+    sr::constants::{Constant, ConstantEnum},
+    sr::types::{StructMember, TypeEnum, Type},
+};
+
+type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
+/// An index corresponding to `Id` in SPIR-V.
+type Index = u32;
+
+/// A strongly typed reference to a SPIR-V element.
+#[derive(Debug)]
+pub struct Token<T> {
+    index: Index,
+    marker: PhantomData<T>,
+}
+
+impl<T> Clone for Token<T> {
+    fn clone(&self) -> Self {
+        Token {
+            index: self.index,
+            marker: self.marker,
+        }
+    }
+}
+impl<T> Copy for Token<T> {}
+impl<T> PartialEq for Token<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.index == other.index
+    }
+}
+impl<T> Eq for Token<T> {}
+
+impl<T> Token<T> {
+    pub(in crate::sr) fn new(index: Index) -> Self {
+        Token {
+            index,
+            marker: PhantomData,
+        }
+    }
+}
+
+/// A structure holding some kind of SPIR-V entity (e.g., type, constant,
+/// instruction, etc.) that can be referenced.
+#[derive(Debug)]
+pub struct Storage<T> {
+    map: FastHashMap<Index, T>,
+    next_id: Index,
+}
+
+impl<T> Storage<T> {
+    fn new() -> Self {
+        Storage {
+            map: FastHashMap::default(),
+            next_id: 0,
+        }
+    }
+
+    /// Associate a value with a given index, returning a typed token.
+    ///
+    /// This is useful when processing a module in the data representation,
+    /// where the indices are known.
+    pub fn assign(&mut self, index: Index, value: T) -> Token<T> {
+        self.next_id = self.next_id.max(index + 1);
+        let old = self.map.insert(index, value);
+        assert!(old.is_none());
+        Token::new(index)
+    }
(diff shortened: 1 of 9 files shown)


rspirv/sr/context.rs
```diff
@@ -12,11 +12,89 @@
...
-use super::{Type, TypeToken, Constant, ConstantToken};
-use sr::constants::ConstantEnum;
-use sr::types::{StructMember, TypeEnum};
+use sr::constants::{Constant, ConstantEnum};
+use sr::types::{StructMember, TypeEnum, Type};
+
+type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
+type Index = u32;
+
+#[derive(Debug)]
```
antiagainst: Can we have documentation for public APIs?

rspirv/sr/context.rs
```diff
@@ -12,11 +12,89 @@
...
+impl<T> Token<T> {
+    pub(in sr) fn new(index: Index) -> Self {
+        Token {
+            index,
+            marker: PhantomData,
+        }
+    }
+}
+
+/// A structure holding some kind of a structured SPIRV data that can be
```
antiagainst: ... some kind of SPIR-V entity (e.g., type, constant, instruction, etc.) that can be referenced.

rspirv/sr/context.rs
```diff
@@ -12,11 +12,89 @@
...
+
+impl<T> Storage<T> {
+    fn new() -> Self {
+        Storage {
+            map: FastHashMap::default(),
+            next_id: 0,
+        }
+    }
+
+    pub fn assign(&mut self, index: Index, value: T) -> Token<T> {
```
antiagainst: Documentation for public API. Same for the following methods.

kvark:
Updated with respect to the review notes (thanks for the quick review!).

I think there is a fatal flaw in this logic though. It's meant to support a scenario where the data representation module is loaded, then the user adds or removes some data, and then saves as a new data module.
The problem is how to associate SPIR-V `IdRef` indices with these internal ones. The PR attempts to make them 1:1 matching (hence the `u32` and `HashMap` instead of a `Vec`). If they don't match, there will need to be some redirection table, which would be unfortunate to keep and update.
So the problem is that the indices need to be unique at the `Context` level, not at the `Storage` level.  We'd need to move `next_id` to the `Context` and somehow use it consistently.

How does this sound to you?

kvark:
@antiagainst I made the `max_index` to be storage-agnostic now. The workflow is expected to be the following:
  1. load a data rep module from a SPIR-V
  2. process into `LoadContext`, can analyze elements in structured way at this point
  3. convert into `Context`, can modify the elements
  4. export into a data rep module
  5. save the data rep into SPIR-V

rspirv/sr/context.rs
```diff
@@ -12,11 +12,128 @@
...
+/// The context accumulated from loading a SPIR-V moduke from the data
+/// representation.
+#[derive(Debug)]
+pub struct LoadContext {
+    /// All type objects.
+    types: Storage<Type>,
+    constants: Storage<Constant>,
+}
+
+impl LoadContext {
```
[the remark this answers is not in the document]
kvark: Yes, see https://github.com/gfx-rs/rspirv/pull/49#issuecomment-513941231
The user creates a `LoadContext`, reads a data-rep module into it, then converts to the `Context`, which can be modified by hands.

kvark:
> I was thinking we may want to completely detach the inside the binary module with the token we use in the memory arena.

So we have an important arhcitectural choice here, and I don't want to rush/force into a particular solution. I want the choice to be well reasoned about :)
  1. Tokens are separate from IdRefs. We'd need some bi-directional mapping set up in order to convert to and from the data representation. Could be a `Map<idRef, TokenIndex>` with each `Token` storing the `IdRef` internally as extra data.
  2. Tokens are exactly the typed IdRefs. No mapping is required, but there is a minor implementation inconvenience for code that adds new tokens, since it has to consider the index to be shared with other tokens.

First important thing to stress is that this choice is *mostly* an implementation detail. The only difference so far is the separate `LoadContext` type in approach (2). The benefit of (2) is that it's easier to convert to/from data representations (basically, can forward-declare tokens), and it may be faster (which is quite important). What would be the benefits of (1) ?

AIOOB:
I think the benefit of (1) is that you can drop that map as soon as you have finished parsing it, as there is no requirement for the ids to be the same if it gets serialised out to SPIR-V. This would also mean it could generate slightly nicer modules on output as the ids can be guaranteed sequential.
As far as I can tell (2) is just (1) with an extra promise (of keeping the ids the same), which I think is shown by the fact that `LoadContext` doesn't actually need to be a seperate type from `Context` as it should be inacessible until parsing is done anyway?
(1) would also support something like #23 more readily as you wouldn't have to worry about ids clashing at all.

antiagainst:
Yep, @AIOOB captured it well; thanks!

I think one thing that's worth mentioning is that using explicit <id> numbers is just SPIR-V's representative choice; it can opt to have each instruction to generate an implicit <id> sequentially (i.e., the instruction index in the binary) and then we can reference the implicit <id>. That will avoid the problem of arbitrary large <id> numbers and eliminate the need for the <id> condensing pass. (And many instructions save a word for encoding the <id>.) But SPIR-V did not choose to do it that way.

I'd like to have the SR to focus more on facilitating transformations, which inevitably means changes. So sticking to the existing <id>s is likely a pain for us to guarantee; so thus I think completely detaching from the <id> in SR is preferable. 

As said by @AIOOB, I think we don't really need the original <id> once done importing from DR, except the case of for better validation error messages.

kvark:
Alright, thanks for feedback!
Closing in favour of #55

PR #53, opened 2019-07-24
Remove most unsafe from the repo

AIOOB:
This currently removes all bar one `unsafe` from the repo and replaces some hand-rolled methods with equivalent ones from `std`.
The remaining `unsafe` is here:
https://github.com/gfx-rs/rspirv/blob/614e9195aac7142b911acd430a521cc069d95a79/rspirv/binary/parser.rs#L204-L208
While I think this is technically safe it will prevent the parser from parsing perfectly legitimate big endian encoded files, as per this bit of the spec:
> Tip
> Endianness: A module is defined as a stream of words, not a stream of bytes. However, if stored as a stream of bytes (e.g., in a file), the magic number can be used to deduce what endianness to apply to convert the byte stream back to a word stream.

It is also doing work twice, converting `u32` slices to `u8` slices then back again in the decoder.  I think it might make sense for the decoder to store a reader or iterator over a trait that is generic over bytes and words, which we would almost certainly have to write, `byteorder` doesn't seem quite up to the task. Being able to take something that imlements `Read` would be nice as then consumers would be able to pass files, etc directly to the API.


rspirv/binary/decoder.rs
```diff
@@ -414,33 +409,29 @@ mod tests {
...
-
-        let b = f32_to_bytes(-12.34);
-        let mut d = Decoder::new(&b);
-        assert_eq!(Ok(-12.34), d.float32());
     }
 
     #[test]
     fn test_decode_float32() {
-        let b = f32_to_bytes(42.42);
+        let b = [0x14, 0xAE, 0x29, 0x42];
```
kvark: would be a little better if there was a conversion from `42.42` here instead of the hard-coded binary

AIOOB: The problem with that is you'll just end up with the same code in the test as is in the code it's testing which seems a bit pointless, but I agree it isn't very nice. It could probably be improved if the decoder accepted words because at least then the input would be in a SPIR-V format rather than just seemingly random bytes.

antiagainst:
Yeah, I'm not sure big endian is still a thing. So didn't actually support it from the very beginning.

PR #54, opened 2019-07-25
Adds required capabilities to operand kinds

AIOOB:
This seemed the most sensible place to put this, I'll happily move it if there is somewhere better. The other piece of information similar to this is the extensions field, but that is currently only on a few types and it seems silly to put it on every one. Also seen as every single one now shares the same function signature maybe there should be a trait linking them up?


AIOOB:
The new change was just changing a `HashMap` to a `BTreeMap` so that the autogen would be stable.

AIOOB:
Rebased and added some line breaks in the output

PR #55, opened 2019-07-25
sr: Generic Storage with Tokens detached from SPIR-V references

kvark:
This is an alternative to #49 that implements `Token` in a way that is independent from SPIR-V references. There is still a bidirectional link in place to allow lookups upon building the structured representation.

Overall, this approach feels cleaner to me, but I do anticipate the structured processing to be more difficult with it, since we aren't able to use any tokens that we haven't processed yet at any point.

cc @AIOOB feedback is welcome!

### DIFF
--- a/rspirv/sr/context.rs
+++ b/rspirv/sr/context.rs
@@ -16,18 +16,24 @@ use std::{
     collections::HashMap,
     hash::BuildHasherDefault,
     marker::PhantomData,
+    ops,
 };
 
 use fxhash::FxHasher;
+use spirv;
 
 use crate::{
-    spirv,
     sr::constants::{Constant, ConstantEnum},
     sr::types::{StructMember, TypeEnum, Type},
 };
 
 type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
-/// An index corresponding to `Id` in SPIR-V.
+/// An unique index in the storage array that a token points to.
+///
+/// This type is independent of `spirv::Word`. `spirv::Word` is used in data
+// representation. It holds a SPIR-V and refers to that instruction. In
+// structured representation, we use Token to refer to an SPIR-V instruction.
+// Index is an implementation detail to Token.
 type Index = u32;
 
 /// A strongly typed reference to a SPIR-V element.
@@ -54,7 +60,13 @@ impl<T> PartialEq for Token<T> {
 impl<T> Eq for Token<T> {}
 
 impl<T> Token<T> {
-    pub(in crate::sr) fn new(index: Index) -> Self {
+    #[cfg(test)]
+    pub const DUMMY: Self = Token {
+        index: !0,
+        marker: PhantomData,
+    };
+
+    fn new(index: Index) -> Self {
         Token {
             index,
             marker: PhantomData,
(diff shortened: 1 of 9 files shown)


rspirv/sr/context.rs
```diff
@@ -54,58 +58,88 @@ impl<T> PartialEq for Token<T> {
...
+
 /// A structure holding some kind of SPIR-V entity (e.g., type, constant,
 /// instruction, etc.) that can be referenced.
 #[derive(Debug)]
 pub struct Storage<T> {
-    map: FastHashMap<Index, T>,
-    next_id: Index,
+    /// Values of this storage.
+    data: Vec<Data<T>>,
+    /// Reverse lookup table that associates SPIR-V references with tokens.
```
antiagainst: Can we be consistent on terms used? It will make the code and comment more consistent and lower the barrier for understanding the code.

We should use `<id>`, which is used in the SPIR-V spec consistently, to mean the "references" in the raw binary and DR. We should use `Token` to "refer" instructions in SR. I think `Index` is just an implementation detail to SR that implies we are using a `Vec` internally to store all the instructions in SR. Public users should not be concerned about that.

rspirv/sr/context.rs
```diff
@@ -16,18 +16,22 @@ use std::{
...
-    spirv,
     sr::constants::{Constant, ConstantEnum},
     sr::types::{StructMember, TypeEnum, Type},
 };
 
 type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
-/// An index corresponding to `Id` in SPIR-V.
+/// An unique index in the storage array that a token points to.
+///
+/// This type is independent of `spirv::Word`.
```
antiagainst: More on the difference? What about:

This type is independent of `spirv::Word`. `spirv::Word` is used in data representation. It holds a SPIR-V <id> and refers to that instruction. In structured representation, we use `Token` to refer to an SPIR-V instruction. `Index` is an implementation detail to `Token`.

rspirv/sr/context.rs
```diff
@@ -54,58 +58,88 @@ impl<T> PartialEq for Token<T> {
...
 
-    /// Associate a value with a given index, returning a typed token.
-    ///
-    /// This is useful when processing a module in the data representation,
-    /// where the indices are known.
-    pub fn assign(&mut self, index: Index, value: T) -> Token<T> {
-        self.next_id = self.next_id.max(index + 1);
-        let old = self.map.insert(index, value);
-        assert!(old.is_none());
+    /// Associates the given value to the given SPIR-V IdRef inside this storage
```
antiagainst: Let's use `<id>` instead of IdRef. `<id>` is the spec's term.

rspirv/sr/context.rs
kvark: I agree about the consistency in terms. Just need to get used to them :)

As for `Index` - it is still private, and this `lookup` field is private.

rspirv/sr/context.rs
```diff
@@ -54,58 +58,88 @@ impl<T> PartialEq for Token<T> {
 impl<T> Eq for Token<T> {}
 
 impl<T> Token<T> {
-    pub(in crate::sr) fn new(index: Index) -> Self {
+    pub const DUMMY: Self = Token {
```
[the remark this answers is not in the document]
kvark: It's for tests only, I'll add `cfg` on top

rspirv/sr/context.rs
```diff
@@ -54,58 +58,88 @@ impl<T> PartialEq for Token<T> {
...
         }
     }
 }
 
+
+/// Generic data that has an associated reference to the source SPIR-V.
+#[derive(Debug)]
+struct Data<T> {
+    value: T,
+    raw_index: Option<NonZeroU32>,
```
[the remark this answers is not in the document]
kvark: I don't clearly see if we'll need that. Let me remove it, and we'll see if it needs to be re-added in some form later.

kvark:
@antiagainst this is now rebased, with comments addressed. Please take another look!

PR #57, opened 2019-07-25
codegen: separate Branch from Terminator

kvark:
In the spec, "Terminator" defined as either a "Branch" or the other two instructions. It would be helpful for the structured representation to work with branch instructions that are terminators.

### DIFF
--- a/codegen/external/README.md
+++ b/codegen/external/README.md
@@ -6,3 +6,6 @@ repository, with some modifications to spirv.core.grammar.json. So there is
 a copy of it in this directory.
 
 [spirv-headers]: https://github.com/KhronosGroup/SPIRV-Headers
+
+Modifications are:
+  - "class" field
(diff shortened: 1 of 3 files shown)


codegen/external/README.md
```diff
@@ -6,3 +6,6 @@ repository, with some modifications to spirv.core.grammar.json. So there is
 a copy of it in this directory.
 
 [spirv-headers]: https://github.com/KhronosGroup/SPIRV-Headers
+
+Modifications are:
+  - "class" field
```
AIOOB: Might as well be a bit more specific here, just in case we end up modifying enough that diffs get confusing if we have to update it.

kvark: The thing is, my PR doesn't introduce the divergence, it's already there.
I just left this as a beginning of some kind of a doc to reflect that. I can remove this readme change if you think it would be nicer, I don't think it's harmful though.

AIOOB: Fair enough, I didn't realise it had already been modified.

antiagainst: Yeah, I introduced the modification to facilitate code autogen. Should really consider upstream this. It's a pain to maintain.

PR #61, opened 2019-07-28
Derive PartialEq, Debug and Clone on public types.

icefoxen:
Some of them are probably unnecessary, and there's probably some I missed.  But it's convenient to be able to wrap rspirv's types in my own types and then derive Debug on 'em.

### DIFF
--- a/codegen/binary.rs
+++ b/codegen/binary.rs
@@ -57,7 +57,7 @@ pub fn gen_operand_decode_errors(grammar: &Vec<structs::OperandKind>)
     }).collect();
     let error_enum = format!(
         "/// Decoder Error.\n\
-         #[derive(Debug, PartialEq)]\n\
+         #[derive(Debug, PartialEq, Clone)]\n\
          pub enum Error {{\n\
          {s:4}StreamExpected(usize),\n\
          {s:4}LimitReached(usize),\n\
--- a/rspirv/binary/autogen_error.rs
+++ b/rspirv/binary/autogen_error.rs
@@ -22,7 +22,7 @@ use spirv;
 use std::{error, fmt};
 
 /// Decoder Error.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub enum Error {
     StreamExpected(usize),
     LimitReached(usize),
--- a/rspirv/binary/decoder.rs
+++ b/rspirv/binary/decoder.rs
@@ -75,6 +75,7 @@ const WORD_NUM_BYTES: usize = 4;
 ///     assert_eq!(Err(DecodeError::StreamExpected(12)), d.word());
 /// }
 /// ```
+#[derive(Clone, Debug, PartialEq)]
 pub struct Decoder<'a> {
     /// Raw bytes to decode
     bytes: &'a [u8],
--- a/rspirv/binary/tracker.rs
+++ b/rspirv/binary/tracker.rs
@@ -35,7 +35,7 @@ pub enum Type {
 ///
 /// If the type of an id cannot be resolved due to some reason, this will
 /// silently ignore that id instead of erroring out.
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct TypeTracker {
     /// Mapping from an id to its type.
     ///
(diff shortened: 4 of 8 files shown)


icefoxen:
If we can derive something, then *obviously* we should at every opportunity!  Right?  :grin: 

It breaks nothing for me either way, so.

kvark:
> If we can derive something, then obviously we should at every opportunity! Right?

No, we should think of what needs to be derived.
All `Copy` gives you is hiding the cost of a copy. It's no longer explicit from the programmer side. If you have a big object, you don't want to waste more time copying it around in memory, even if Rust sees it as `Copy`-compatible.

icefoxen:
For me, what needs to be derived is "whatever makes it convenient to unit test things easily", but that wasn't something that was thought of.  And the cost of copying a `ModuleHeader` is 20 bytes, which seems pretty small.  So I tend to favor the permissive approach.

antiagainst:
I don't quite see the benefits of making everything copyable, esp. `Decoder`, `Builder`, etc. Those classes are intended to maintain some context and update its internal state gradually to eventually derive something valid and meaningful; at a certain time during the process its internal stage may be invalid or not meaningful to "take a snapshot" by copying. I don't quite see the purpose of enabling copying of these objects given they are meant as the sole object for decoding a certain binary blob or building for a certain SPIR-V module. I don't think we should go down the route of enabling copyable for all structs simply because we can or it simplifies the tests. What we enable will be come public APIs and there may be dependencies relying on the feature; it would be harder to take back once there. (Technically rspirv is still pre-1 stage; but still we don't want to always break, right? ;-P)

What's the intended use cases for enabling these `Copy` traits? Could you give examples?

icefoxen:
`Copy` I don't need but `Clone` I do.  The use case is simplifying the test cases.  I would also expect cloning a `Builder` to be faster than re-building one.

For that matter it would be nice if I could, from time to time, make copies to inspect the SPIR-V they contain and make sure it matches what I am trying to generate; this is useful for debugging-via-unit-test as well as basic validation.  For example, add a function, assert it is correct, add another one, assert that it calls the first one correctly, and so on.  You explain in the docs that the innards of the builder does not necessarily contain well-formed SPIR-V, and that's okay.  I'm trying to make tests that inspect the code generation process piece by piece so if something unexpectedly changes I notice early instead of late.

I'm writing a compiler.  It's complicated and error prone.  Simplifying the test cases and allowing incremental development is useful.

That said, some things such as `Decoder` I threw those traits on just 'cause I was trying to be thorough.  I do maintain that there is no good reason to ever not implement `Debug` on anything though, unless you have binary size constraints.

icefoxen:
All right, let's agree to disagree for now, 'cause I don't want this PR in limbo forever. Which types should I remove the clone derive's for, just Decoder and Builder?

antiagainst:
I understand your use case better now. Thanks for the explanation!

Would you mind to explain what compiler you are building if possible? I feel it might be better for you to use the structured representation that we are developing right now (which saves you from some low-level boilerplate), but I'm not sure about your schedules and plans.

Instead of marking Builder/Decoder/etc. as `Copy`/`Clone`, if you just want to inspect the internals of a Builder, what do you think if we add a method to Builder, like `peek_module` to give you a handle to the module temporarily? I guess it's simpler for you since you don't need to clone always.

icefoxen:
The compiler is here: https://hg.sr.ht/~icefox/chrysanthemum  .  It's very young and not really ever intended to be complete, but hopefully it can at least help serve as a demo for other people who are interested in such things.  So, honestly, any suggestions for how to use `rspirv` better would be useful.  A structured representation would definitely be interesting but honestly building mostly-raw SPIR-V is fairly compiler-friendly so far and I'm happy with just not have to worry about assembling and outputting it correctly.

Actually, adding a `peek_module()` method or such would be just fine; if it doesn't take much work to finalize then returning `&Module` would be perfect.

antiagainst:
Cool stuff! Thanks for the pointers!

I'd happy with `peek_module()` if it works for you. So please feel free to proceed with a PR to add it. :)

antiagainst:
@icefoxen: I'll close this one now. Feel free to have another PR to add `peek_module` as discussed above. Thanks! :)

PR #63, opened 2019-07-29
Structured module items

kvark:
This is another small step towards #5 , taken out of #46 
It's based on #62, so only the second commit needs to be reviewed here.
We are now generating a few enums and a few standalone structs for the instructions.

### DIFF
--- a/rspirv/sr/autogen_instruction_structs.rs
+++ b/rspirv/sr/autogen_instruction_structs.rs
@@ -1,0 +1,63 @@
+// Copyright 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// AUTOMATICALLY GENERATED from the SPIR-V JSON grammar:
+//   external/spirv.core.grammar.json.
+// DO NOT MODIFY!
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Extension {
+    name: String,
+}
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ExtInstImport {
+    name: String,
+}
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct MemoryModel {
+    addressing_model: spirv::AddressingModel,
+    memory_model: spirv::MemoryModel,
+}
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct EntryPoint {
+    execution_model: spirv::ExecutionModel,
+    entry_point: spirv::Word,
+    name: String,
+    interface: Vec<spirv::Word>,
+}
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ExecutionMode {
+    entry_point: spirv::Word,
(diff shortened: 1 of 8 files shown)


antiagainst:
This needs a rebase now. :)

kvark:
Rebased now

rspirv/sr/autogen_instruction_structs.rs
```diff
@@ -0,0 +1,63 @@
...
+pub struct Capability {
+    capability: spirv::Capability,
+}
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Function {
+    function_control: spirv::FunctionControl,
+    function_type: spirv::Word,
+}
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct FunctionParameter {}
```
antiagainst: Do we still want to have separate ops for these function components? I think they should be folded into the function SR itself. I mean, `FunctionEnd` is useful for marking the end of a function in SPIR-V binary, but not useful for SR. Similarly for `Label`, it should be implicit for BasicBlock SR. Letting each pass processing Function or BasicBlock to worry about `FunctionEnd` and `Label` is not helpful.

rspirv/sr/items.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub capabilities: Vec<op::Capability>,
+    /// All OpExtension instructions.
+    pub extensions: Vec<op::Extension>,
+    /// All OpExtInstImport instructions.
+    pub ext_inst_imports: Vec<op::ExtInstImport>,
+    /// The OpMemoryModel instruction.
+    ///
+    /// Although it is required by the specification to appear exactly once
+    /// per module, we keep it optional here to allow flexibility.
+    pub memory_model: Option<op::MemoryModel>,
```
antiagainst: Do we want to still keep it optional in SR? I think we should mandate it here.

rspirv/sr/items.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub result: Token<Type>,
+    /// Function parameters.
+    pub parameters: Vec<Token<Type>>,
+    /// Basic blocks in this function.
+    pub basic_blocks: Vec<BasicBlock>,
+}
+
+pub struct Module {
+    /// The module header.
+    pub header: Option<ModuleHeader>,
```
antiagainst: This should be mandated?

rspirv/sr/items.rs
```diff
@@ -0,0 +1,53 @@
...
+use spirv;
+
+pub struct EntryPoint {
+    pub execution_model: spirv::ExecutionModel,
+    pub entry_point: Token<Function>,
+    pub name: String,
+    //pub interface: Vec<spirv::Word>,
+}
+
+pub struct BasicBlock {
```
antiagainst: Where is the Vector for instructions?

rspirv/sr/autogen_instruction_structs.rs
kvark: I agree, we'll need to set up a list of ops that don't need SR autogen generation, and which we are going to be converting to and from DR manually. We'll do it as a follow-up.

kvark:
Thanks for the review! All the concerns are addressed now, I believe.

kvark:
I think the commit message is descriptive enough? If not, let's talk about this some more and figure out good guidelines.

antiagainst:
I typically prefer to have a concise onliner for describing what the commit is doing. The onliner shouldn't be too long so it renders nicely on GitHub UI (otherwise we will see ... and broken sentences) and also CLI log dump.

Then have some more detailed explanation following after an empty line. I'd prefer to have the detailed explanation to have proper line breaks so each paragraph is not on a single line. (Otherwise when we do `git log`, it's quite annoying to have extra long lines, especially we you use `tmux` or `screen` and are in a small tab, then the message will just be truncated.)

antiagainst:
The content of the message is good (modulo a typo for `module` but not a big thing)! Thanks! :)

PR #64, opened 2019-07-29
Codegen refactor

kvark:
The less macros the better!

### DIFF
--- a/codegen/build.rs
+++ b/codegen/build.rs
@@ -20,42 +20,41 @@ mod structs;
 mod table;
 mod utils;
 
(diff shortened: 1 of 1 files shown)


antiagainst:
Could you add more explanations of what the commit is doing in the message in the future? The commit message will be come a part of the history; it would be nice to be clear on what a commit is doing without reading the implementation. A commit just with a "codgen refactor" message does not help much in understanding the commit. :)

kvark:
Agreed, this isn't the best commit message. Thanks for the heads up!

PR #66, opened 2019-08-02
Rename codegen to autogen, remove it from rspirv dependencies

kvark:


antiagainst:
Thanks for taking care of this!

Issue #42, opened 2019-07-15
Case against is_type() methods


kvark:
There is a large auto-generated section looking like:
```rust
impl Type {
  pub fn is_foo(&self) -> bool {...}
  pub fn is_bar(&self) -> bool {...}
  ...
}
```
It enumerates each and every enum variant.

What is the motivation behind this API? I think we could live without it. The more complex (and semantically rich) checks would stay and just pattern match internally:
```rust
impl Type {
    pub fn is_numerical_type(&self) -> bool {
    }
    pub fn is_scalar_type(&self) -> bool {
    }
    pub fn is_aggregate_type(&self) -> bool {
    }
    pub fn is_composite_type(&self) -> bool {
    }
}
```

antiagainst:
The rationale is that in compiler code base checking a type's value is quite prevalent. Just trying to throw out some line savings here by turning the 4-line `match` into one method call. It's small-ish per site, but it adds up. :)

Given this is already there; we can keep it. After coding more components out then we can re-evaluate the situation and delete it if it ends up not being used that much.

kvark:
Continuing from https://github.com/gfx-rs/rspirv/pull/67#issuecomment-518730528

> Hiding internals of Type and make its new method not public is for that purpose.

I don't think this conflicts with the idea of having `Context` to de-duplicate the types. You can still do this:
```rust
let type = Type { ... }; // set it up manually
let token = context.add_type(type);
```

> probing its properties, which is intended to be achieved via those is_* method

But `is_*` method doesn't *give* you the properties, it just tells you if a type belongs to a particular kind. What if, say, I want to analyze SPIR-V types and map them to Metal types. I got `Type { ty: TypeEnum::Int { width, signedness }, decorations: Vec::new() }`. If all I can do is `is_int()` on that type, how do I figure out if it maps to `int`, `uint`, `short`, etc?

What is the downside of just exposing the type system instead of hiding it behind a bunch of `new_xx` and `is_xxx` methods?

antiagainst:
I think we are conflating multiple things here:
1. Whether having `is_*_type` checks is useful (which is what this issue about)
1. Whether we want to make Type construction only available via `Context`
1. Whether we want to expose the internals of `Type`
1. Whether we want to merge `Context` into `Module`

Let's discuss one at each time instead of mixing them together. :)

For 1.

[...]

PR #67, opened 2019-08-02
Centralized SR code generation

kvark:


kvark:
@antiagainst I think it's looking more comprehensible now, saving about 40 LOC on the way :) Please take a look!

rspirv/sr/autogen_type_enum_check.rs
```diff
@@ -17,7 +17,7 @@
 // DO NOT MODIFY!
 
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub(in crate::sr) enum TypeEnum {
+pub enum TypeEnum {
```
antiagainst: This should be `Type` internal; why making it entirely public?

rspirv/sr/types.rs
```diff
@@ -29,7 +29,7 @@ pub struct Type {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub(in crate::sr) struct StructMember {
+pub struct StructMember {
```
antiagainst: Same here. Why leaking this symbol to the public?

autogen/sr.rs
```diff
@@ -114,166 +74,70 @@ fn get_type_fn_name(name: &str) -> String {
...
-pub fn gen_sr_type_check(grammar: &structs::Grammar) -> String {
-    // Collect all types and their parameters in the following format:
-    //   (type-name: &str, Vec<(param-name: quote::Ident, param-type: quote::Ident)>)
-    let cases: Vec<_> = grammar
-        .instructions
+pub struct GeneratedOperands {
+    pub decoration: String,
+}
+
+pub fn gen_sr_operands(grammar_operand_kinds: &[structs::OperandKind]) -> GeneratedOperands {
```
antiagainst: Can we name this function as `gen_sr_code_from_operand_kind_grammar` or something? "Operand" is already fairly overloaded; let's differentiate wherever possible.

autogen/sr.rs
```diff
@@ -114,166 +74,70 @@ fn get_type_fn_name(name: &str) -> String {
     }
 }
 
-pub fn gen_sr_type_check(grammar: &structs::Grammar) -> String {
-    // Collect all types and their parameters in the following format:
-    //   (type-name: &str, Vec<(param-name: quote::Ident, param-type: quote::Ident)>)
-    let cases: Vec<_> = grammar
-        .instructions
+pub struct GeneratedOperands {
```
antiagainst: Name it as `CodeGeneratedFromOperandKindGrammar` or similar?

autogen/sr.rs
```diff
@@ -114,166 +74,70 @@ fn get_type_fn_name(name: &str) -> String {
...
-            }
-        })
-        .collect();
-    let tokens = quote! {
-        impl Context {
-            #( #constructors )*
-        }
-    };
-    tokens.to_string()
+pub struct GeneratedInstructions {
```
antiagainst: Name it as `CodeGeneratedFromInstructionGrammar` or similar?

autogen/sr.rs
```diff
@@ -114,166 +74,70 @@ fn get_type_fn_name(name: &str) -> String {
...
+    pub type_enums: String,
+    pub type_creation: String,
+    pub instruction_structs: String,
+    pub instruction_enums: String,
 }
 
-/// Generates instruction enums and standalone structs.
-pub fn gen_sr_instruction(grammar: &structs::Grammar) -> (String, String) {
-    let mut structs = Vec::new();
+pub fn gen_sr_instructions(grammar_instructions: &[structs::Instruction]) -> GeneratedInstructions {
```
antiagainst: Name it as `gen_sr_code_from_instruction_grammar` ?

autogen/sr.rs
```diff
@@ -292,13 +156,89 @@ pub fn gen_sr_instruction(grammar: &structs::Grammar) -> (String, String) {
...
+                    })
+                    .collect();
+                let param_list = param_list.as_slice();
+                type_variants.push(if param_list.is_empty() {
+                    quote!{ #symbol }
+                } else {
+                    quote! { #symbol { #( #param_list ),* } }
+                });
+
+                if true {
```
antiagainst: Why the `if true` here?

kvark:
Thanks for the review!

I renamed the types and functions as advised.
As for the public access, how else would the user be able to inspect SR of SPIR-V? My understanding was that we want to expose a nice type layer where users would match on their own and construct new data. That's also why I don't think `is_xxx` accessors are needed, by the way.

antiagainst:
I actually view the `TypeEnum` field as internal to `Type`. All types should be uniqued into `Context` and created from `Context`. Hiding internals of `Type` and make its `new` method not public is for that purpose. There are two asks regarding `Type`: probing its properties, which is intended to be achieved via those `is_*` method, and constructing new types, which should be done via the `Context`. Without opening the internals of `Type` to the public, I don't quite see why making `TypeEnum` public would be useful? Instead of letting each call site to have a match construct to probe what kind a type is, why `is_*` not a nicer API that saves the boilerplate? 

Anyway, I think the change to publicize `TypeEnum` should be in another PR given that is not relevant to the PR here (which is trying to refactoring autogen) and should be discussed separately. Can you undo it to have this PR submitted first?

antiagainst:
Typically I think smaller PRs dedicated to one task would be nice for review and revert if something goes wrong later.

kvark:
Thanks for the review!
I agree the public qualifier change needs to be discussed separately (responded in #42). It is now reverted here. Are we good to go?

PR #73, opened 2019-08-07
Integration test for external SPIR-V modules.

### LINKED ISSUE #70: Testing with real SPIR-V data
We need a mechanism for testing (locally and on CI) the binary parser and DR -> SR conversion on some real data. Preferably, without including any actual SPIR-V modules in the repository, as requested by @antiagainst .

kvark:
Fixes #70

The test scans through the "test-data" directory tree and searches for "*.spv"
files, treating them as valid SPIR-V modules. It attemps to parse them for
the data representation and assemble/disassemble the module.


rspirv/tests/test_data.rs
```diff
@@ -0,0 +1,48 @@
...
+                    if ext.to_string_lossy() != "spv" {
+                        continue
+                    }
+                }
+                None => continue
+            }
+            let spv = fs::read(path).unwrap();
+            let module = dr::load_bytes(spv).unwrap();
+            let _disasm = module.disassemble();
+            let _assembly = module.assemble();
```
AIOOB: I am slightly worried some of this could get optimized away, although I'm not sure how this could be addressed.

rspirv/tests/test_data.rs
```diff
@@ -0,0 +1,48 @@
+use rspirv::{
+    binary::{Assemble as _, Disassemble as _},
```
AIOOB: Does this serve a purpose? I think it's better not to have the `as` if it doesn't.

kvark: I find this to be a good way to use what the traits offer *without* including the trait names themselves into the visible space. I.e. it's a more clear indication of the intent here.

kvark:
@antiagainst @AIOOB thanks for the reviews!
comments are addressed, please take another look

PR #74, opened 2019-08-07
Change remaining autogen to use tokens

### LINKED ISSUE #45: Improve codegen grammar typing
The biggest gain would be having strongly typed parsing results, e.g.
`if let Class::ModeSetting = inst.class  { ... }` instead of `if inst.class = "ModeSetting" { ... }`

We should evaluate the cost bringing the dependency, the amount of code removed from here, and the benefits.

### LINKED ISSUE #47: Port "mr" generators to tokens
The whole "codegen/mr.rs" is implemented to emit strings. This doesn't look nearly as tidy as `quote!`-based workflow of "codegen/sr.rs". We should make those consistent.

AIOOB:
Fixes #47
Fixes #45


AIOOB:
The changes to the deserialisation are included here as there were a few mistakes caused by the struct that should have been an enum. It seemed simpler to include the fix for the root cause than search for the bugs.

kvark:
This is some amazing work! You might want to check #75 and restructure these changes into a number of PRs though, accordingly.

PR #75, opened 2019-08-08
Update CONTRIBUTING.md to list pull request and commit conventions

antiagainst:
These conventions will help us to speed up code review process
and keep a clean and healthy Git history for future reference.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,27 +1,40 @@
-Want to contribute? Great! First, read this page (including the small print at the end).
+# Contributing to rspirv
 
-### Before you contribute
-Before we can use your code, you must sign the
-[Google Individual Contributor License Agreement]
-(https://cla.developers.google.com/about/google-individual)
-(CLA), which you can do online. The CLA is necessary mainly because you own the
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,27 +1,40 @@
...
-(https://cla.developers.google.com/about/google-corporate).
+# Contributing to rspirv
+
+It's great to know that you are looking to contribute to rspirv! To keep the
+project in a consistent state on various aspects, we have the following
+conventions that it would be very nice if you can consider:
+
+## Code review
+
+* Please address comments with a new commit instead of force pushing to
```
kvark: TIL :)

kvark:
Did you consider to move the `CONTRIBUTING.md` to `.github` folder (like in https://github.com/gfx-rs/gfx/tree/master/.github) so that the root dir is not polluted?

antiagainst:
This is not specific to GitHub; so I think we should just keep it under root. Issue/pull request templates should be under `.github/`. :)

kvark:
It does actually have github-specific treatment- https://help.github.com/en/articles/setting-guidelines-for-repository-contributors

PR #79, opened 2019-08-08
Change header autogen to use tokens

AIOOB:


antiagainst:
Can we format this better?

AIOOB:
I'm not sure but I think the answer is no, rustfmt doesn't seem to be able to cope with bitflags although I'm not sure why it's supposed to format within macro blocks if it looks like rust code, but I couldn't get even that to work.

antiagainst:
I see. SGTM. Let's submit this and figure out the formatting later. :)

PR #80, opened 2019-08-08
Add more types to grammar deserialiser

### LINKED ISSUE #45: Improve codegen grammar typing

AIOOB:
Fixes #45 
This adds a few enums to the deserialiser so that there is less string checking in the autogen code. It also changes the Enumerant value field to be a u32 by parsing the string form, which eliminates the weird struct-that's-actually-an-enum.

### DIFF
--- a/autogen/structs.rs
+++ b/autogen/structs.rs
@@ -16,7 +16,7 @@
 
 use serde::de;
 use serde_derive::*;
-use std::{fmt, result, str};
+use std::{convert::TryInto, fmt, result, str};
 
 #[derive(Debug, Deserialize)]
 pub struct Operand {
@@ -24,13 +24,12 @@ pub struct Operand {
     #[serde(default)]
     pub name: String,
     #[serde(default)]
-    pub quantifier: String,
+    pub quantifier: Quantifier,
 }
 
 #[derive(Debug, Deserialize)]
 pub struct Instruction {
-    #[serde(default)]
-    pub class: String,
+    pub class: Option<Class>,
     pub opname: String,
     pub opcode: u32,
     #[serde(default)]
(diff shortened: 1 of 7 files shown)


spirv/autogen_spirv.rs
```diff
@@ -17,112 +17,112 @@
 // DO NOT MODIFY!
 
 pub type Word = u32;
-pub const MAGIC_NUMBER: u32 = 0x07230203;
+pub const MAGIC_NUMBER: u32 = 119734787;
```
kvark: we shouldn't change the hex output in this module, a regression?

antiagainst:
Nee rebase

spirv/autogen_spirv.rs
AIOOB: That sent me down a rabbit hole but it's back to hex now. `proc_macro2` really doesn't like hex literals.

autogen/structs.rs
```diff
@@ -16,21 +16,20 @@
...
     #[serde(default)]
-    pub quantifier: String,
+    pub quantifier: Quantifier,
 }
 
 #[derive(Debug, Deserialize)]
 pub struct Instruction {
-    #[serde(default)]
-    pub class: String,
+    pub class: Option<Class>,
```
AIOOB: I'm not sure whether an option is best here or whether it would be better to just add a `Normal` class.

antiagainst: Given this is just autogen internal, both are fine for me. :)

PR #82, opened 2019-08-08
SR auto-generated modules refactor

kvark:
This PR attempts to organize the auto-generated modules in a way that
is compatible with the emerging SR logic. All the OpType derived types
are put into "autogen_types". All the context logic is put into
"autogen_context". All the instructions that are meant to be operations
on data (come without a class assigned) are put into "autogen_ops". All
the rest are put into "autogen_instructions".


AIOOB:
It's not immediately clear, is this entirely just a renaming?

kvark:
Yes, this is just renaming and reorganizing the autogen targets. We discussed it briefly and thought this scheme would serve us for a while.

PR #83, opened 2019-08-08
Address most clippy lints

AIOOB:
General cleanup, removes some manual bit twiddling, and silences some extraneous warnings.


antiagainst:
Need rebase

PR #85, opened 2019-08-09
Rename some remaining mr to dr

AIOOB:
Nothing exciting, some functions in autogen were missed in the renaming.


kvark:
I think this is non-controversial, merging.

PR #86, opened 2019-08-14
Lifting function types

kvark:
Based on #71

### DIFF
--- a/autogen/sr.rs
+++ b/autogen/sr.rs
@@ -46,7 +46,7 @@ impl OperandTokens {
                 ),
                 "Field Types" => (
                     quote! { StructMember },
-                    quote! { super::types::StructMember::new(value.clone()) },
+                    quote! { types::StructMember::new(value.clone()) },
                 ),
                 "Parameter Types" => (
                     quote! { Token<Type> },
@@ -57,11 +57,13 @@ impl OperandTokens {
                     quote! { self.types.lookup(*value).unwrap() },
                 ),
                 _ => (
+                    //TODO: proper `Token<>`
                     quote! { spirv::Word },
                     quote! { *value },
                 ),
             },
             "IdMemorySemantics" | "IdScope" | "IdResult" => (
+                //TODO: proper `Token<>`
                 quote! { spirv::Word },
                 quote! { *value },
             ),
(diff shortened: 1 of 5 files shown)


kvark:
@antiagainst FYI, CI is green on all three commits here. So if the code is good, we can merge the PR without squashing and avoid some churn

autogen/sr.rs
```diff
@@ -243,6 +248,12 @@ pub fn gen_sr_code_from_instruction_grammar(
         // Get the token for its enumerant
         let inst_name = &inst.opname[2..];
         let name_ident = Ident::new(inst_name, Span::call_site());
+        let type_name = if inst.opname.len() > 6 {
```
antiagainst: What is this mysterious magic number here? Could you add some comment to explain? Thanks!

rspirv/sr/autogen_ops.rs
```diff
@@ -18,9 +18,31 @@
 
 use crate::sr::{Token, Type};
 #[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Branch {
+    Branch {
+        target_label: spirv::Word,
```
antiagainst: So `spirv:Word` is the DR's way to "link" instructions. Here I think target_label should be `Token<BasicBlock>`.

rspirv/sr/autogen_ops.rs
```diff
@@ -18,9 +18,31 @@
...
+    },
+    BranchConditional {
+        condition: spirv::Word,
```
antiagainst: Here `condition` should be `Token<Instruction>` and labels should be `Token<BasicBlock>`.

If you intend to change it later, would be nice at least to have some TODO in the source code. The existing implementation seems wrong to me.

rspirv/sr/autogen_ops.rs
```diff
@@ -18,9 +18,31 @@
...
+        true_label: spirv::Word,
+        false_label: spirv::Word,
+        branch_weights: Vec<u32>,
+    },
+    Switch {
+        selector: spirv::Word,
+        default: spirv::Word,
+        target: Vec<(u32, Token<Type>)>,
```
antiagainst: Why `Token<Type>`? Shouldn't it be `Token<BasicBlock>`?

rspirv/sr/autogen_ops.rs
```diff
@@ -18,9 +18,31 @@
...
+    },
+    Return,
+    ReturnValue {
+        value: spirv::Word,
```
antiagainst: Token<Instruction>

kvark:
@antiagainst this is rebased and reduced in scope, please take a look.

antiagainst:
Smaller ones are much easier to review! :)

PR #87, opened 2019-08-14
(first steps of) SR module construction from DR

kvark:
This is based on #86 (which is based on #71 ).


antiagainst:
Is this meant to be reviewed? Would be nice to address the conflicts if so. :)

rspirv/sr/items.rs
```diff
@@ -31,25 +36,120 @@ pub struct Function {
...
 pub struct Module {
     /// The module header.
-    pub header: ModuleHeader,
+    pub header: dr::ModuleHeader,
     /// All OpCapability instructions.
-    pub capabilities: Vec<instructions::Capability>,
+    pub capabilities: Vec<spirv::Capability>,
     /// All OpExtension instructions.
-    pub extensions: Vec<instructions::Extension>,
+    pub extensions: Vec<String>,
```
AIOOB: Why have you changed these? I can see pros and cons to both ways round. Also, if this is changed what should happen to the things in instructions atm?

AIOOB: Seems sensible, just need to be careful that we keep support of unknown extended instruction sets.

kvark:
@AIOOB could you remind me where we stand on the standalone types topic?
  1. needed, and `Function` is standalone
  2. needed, but `Function` is a variant of `Op`
  3. not needed

AIOOB:
From my point of view, they just aren't needed (although I don't care between 2&3 that much, 3 is preferable as it's less code); they artificially limit what we can represent with no real benefit. However, I haven't looked at the code for a while and would happily accept 1 if anyone can show me a strong benefit it provides.

kvark:
@antiagainst this is ready to be reviewed now

PR #90, opened 2019-08-14
Replace "regex" dependency by "heck"

### LINKED ISSUE #89: Optimize autogen build
On my machine, touching `autogen` module comes with a 35 second waiting penalty. It would be very useful for our productivity to make it faster.

kvark:
Fixes #89 

Heck is a specialized crate for casing conversions, and it appears
to be sufficient for our use cases. Using heck brings down our
autogen time by a factor of 15x.


kvark:
Note that there are a few changes that come with it, i.e. "fp_something" instead of "fpsomething". It looks better than it was to me, unless there was a specific reason to glue "fp_" before.

kvark:
cc @withoutboats - this change brings down my build script times from 35s to 2.4s (factor of 15x). Amazing work, thank you for the library!

PR #92, opened 2019-08-16
Branch and Terminator enums

kvark:
This is the second half of #86 with the review notes addressed.
"//TODO" comments don't show up in the autogenerated code, I only put them into the autogen itself. We'll need to figure out what to do with `Context` before turning them into real `Token<BasicBlock>` and such (since today context doesn't know about basic blocks).

### DIFF
--- a/autogen/sr.rs
+++ b/autogen/sr.rs
@@ -38,97 +38,122 @@ impl OperandTokens {
         let name = get_param_name(operand);
         let iter = Ident::new(OPERAND_ITER, Span::call_site());
 
-        let (ty, lift_value) = match operand.kind.as_str() {
-            "IdRef" => match operand.name.trim_matches('\'') {
-                "Length" => (
-                    quote! { Token<Constant> },
-                    quote! { self.constants.lookup(*value).unwrap() },
-                ),
-                "Field Types" => (
-                    quote! { StructMember },
-                    quote! { types::StructMember::new(value.clone()) },
-                ),
-                "Parameter Types" => (
-                    quote! { Token<Type> },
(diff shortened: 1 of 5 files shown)


autogen/sr.rs
```diff
@@ -38,97 +38,122 @@ impl OperandTokens {
...
+                    ),
+                    "Parameter Types" => (
+                        quote! { Token<Type> },
+                        quote! { self.types.lookup(*value).unwrap() },
+                    ),
+                    name if name.ends_with(" Type") => (
+                        quote! { Token<Type> },
+                        quote! { self.types.lookup(*value).unwrap() },
+                    ),
+                    //TODO:
```
antiagainst: Super nit: Typically we leave a space between `//` and the following text in comments in C++ world. Is this not true for Rust?

kvark: Typically in Rust we are leaving a space after `//`. However, for special notes that start with TODO, FIXME, and so forth, I'm not putting the space in, which in part indicates that these comments are special. I'm fine with changing this, just explaining the general approach I've been doing in other projects.

kvark:
> But we should consider coupling changes with tests as early as possible once we have enough infra.

As soon as we know that we are on the right path, I'll be happy to add more tests.

kvark:
Going to wait for #93 and #94 to be reviewed first, in case they need to be merged - I don't ant to cause @AIOOB any rebase headaches :)

autogen/sr.rs
antiagainst: Sure. I don't care that much on this.

kvark:
This was reviewed and approved. Rebasing didn't involve any structural changes. Merging.

PR #93, opened 2019-08-16
sr: Remove Type and Constant wrappers

### LINKED ISSUE #42: Case against is_type() methods

AIOOB:
Promotes the type and constant enums to public allowing simpler matching.
Closes #42 
Another debatable PR :grin:, this removes the wrappers to allow users to easily access the data inside and pattern match.

It might be useful to scatter some `#[non_exhaustive]`s around, depending on how forward-compatible we want to be.


rspirv/sr/autogen_instructions.rs
```diff
@@ -46,16 +46,16 @@ pub struct ExecutionMode {
 pub struct Capability {
     pub(in crate::sr) capability: spirv::Capability,
 }
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug)]
 pub struct Function {
```
AIOOB: It doesn't really make sense to compare function instructions for equality, as function are only equal if they are the same bit of code (although even then they can be different if entered via different entry points)

rspirv/sr/autogen_types.rs
```diff
@@ -16,13 +16,8 @@
 //   external/spirv.core.grammar.json.
 // DO NOT MODIFY!
 
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct Function {
```
AIOOB: Technically the spec allows you to put references to functions into composite types, however, the entire CFG is static and it is impossible to call them so it's completely useless as far as I can tell (and probably unsupported by drivers). This means we could leave Function outside the Type enum, as long as we are reasonably confident no-one is going to complain about that omission and that the spec won't change to accommodate that, which we probably can be because it would be a huge change to GPU architectures to allow that.

rspirv/sr/constants.rs
```diff
@@ -14,102 +14,21 @@
...
-}
-
-#[derive(Clone, Debug, PartialEq)]
-pub(in crate::sr) enum ConstantEnum {
+pub enum Constant {
     Bool(bool),
-    I32(i32),
-    U32(u32),
-    F32(f32),
+    // TODO: Change to arbitrary width datatype
```
AIOOB: There aren't any arbitrary sized floating-point libraries on crates as far as I can tell, so we may end up having to roll our own. We could just make the data opaque but people will almost certainly want to be able to modify it.

rspirv/sr/constants.rs
```diff
@@ -14,102 +14,21 @@
...
-    SpecU32(u32),
-    SpecF32(f32),
+    SpecInteger(u32),
+    SpecFloat(f32),
     SpecComposite(Vec<Token<Constant>>),
     SpecOp(spirv::Op, Vec<Token<Constant>>),
 }
-
-impl Constant {
-    pub fn is_bool_constant(&self) -> bool {
```
AIOOB: These methods are actually probably useful, I was getting a bit delete-happy at this point...

rspirv/sr/types.rs
```diff
@@ -16,262 +16,54 @@ use crate::spirv;
...
-    pub(in crate::sr) decorations: Vec<Decoration>,
-}
+include!("autogen_types.rs");
 
-impl StructMember {
-    pub(in crate::sr) fn new(token: Token<Type>) -> Self {
-        StructMember {
-            token,
-            decorations: Vec::new(),
+impl PartialEq for Type {
```
AIOOB: The spec is really dodgy here, Section 2.8, page 22 states:
>Two different type <id>s form, by definition, two different types. It is valid to declare multiple aggregate type <id>s
having the same opcode and operands. This is to allow multiple instances of aggregate types with the same structure to be decorated differently. (Different decorations are not required; two different aggregate type <id>s are allowed to have identical declarations and decorations, and will still be two different types.) Non-aggregate types are different: It is invalid to declare multiple type <id>s for the same scalar, vector, or matrix type. That is, non-aggregate type declarations must all have different opcodes or operands. (Note that non-aggregate types cannot be decorated in ways that affect their type.)

This makes no mention of opaque types, I assume they also are never equal due to their opaqueness but I'm not sure.

antiagainst:
SGTM to make `Type` and `Constant` internals public. But I guess I still have questions regarding deleting the `is_*_type` helper methods and allowing creating `Type` and `Constant` without a `Context`. Since this is pointing to #42, let's have some discussion there first.

AIOOB:
See #42 for discussion.

kvark:
Let's proceed!

PR #94, opened 2019-08-16
sr: Merge Context into Module

AIOOB:
Context and Module were one-to-one so removed Context in favour of Module.
This can probably be arranged better, I'm not sure what the current logic is for the module boundaries.
The Module struct needs to carry around type and constant information so these two structs would have to be constantly paired anyway, this just merges the two.


AIOOB:
See #42 for discussion.

kvark:
Thank you for updating!
Could we move all the `Storage` into a dedicated "storage.rs", perhaps? That would make "items.rs" less clunky.

antiagainst:
+1 to what @kvark said. Not related to this PR, but it would also be nice to rename `items.rs` into something more self-descriptive, like `module.rs` (given it's mostly data structures for module) or something, in a later PR.

kvark:
Great suggestion! And I think it makes sense to do in this PR to minimize the churn of code moving.

PR #100, opened 2019-08-29
Replaces num with num-traits

AIOOB:
This removes unnecessary num components, speeding up compilation.


kvark:
err, @antiagainst why do we have `Cargo.lock` in the tree?

PR #101, opened 2019-09-04
Separate module for lifting infrastructure

kvark:


autogen/build.rs
```diff
@@ -192,7 +192,7 @@ fn main() {
         instructions.ops,
     );
     write_formatted(
-        &autogen_src_dir.join("../rspirv/sr/autogen_module.rs"),
-        instructions.module_logic,
+        &autogen_src_dir.join("../rspirv/lift/autogen_context.rs"),
```
antiagainst: Shouldn't we call this as `autogen_lift_context.rs` to be clear then?

kvark: I had it called this for some time but then figured that "lift" is already in the path, so there is no need to repeat ourselves here. I don't mind either way :)
@AIOOB - do you have a preference?

autogen/sr.rs
```diff
@@ -104,14 +104,14 @@ impl OperandTokens {
             ),
             "PairLiteralIntegerIdRef" => (
                 //TODO: Token<BasicBlock>
```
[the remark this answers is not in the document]
kvark: indeed!

kvark:
Ok, I'm open to revisiting this decision. Currently, the autogen files sometimes filed as the types, sometimes not (when there are many types in a file).

kvark:
Oh, I didn't push the fix to remove the TODO. Sorry about that!

PR #105, opened 2019-09-23
[lift] LiftStorage wrapper

kvark:
Based on #104

### DIFF
--- a/rspirv/lift/storage.rs
+++ b/rspirv/lift/storage.rs
@@ -1,0 +1,61 @@
+use crate::{
+    sr::{
+        storage::{Storage, Token},
+    },
+};
+
+use std::{
+    borrow::Borrow,
+    collections::hash_map::{Entry, VacantEntry, HashMap},
+    hash::BuildHasherDefault,
+};
+
+use fxhash::FxHasher;
+
+
+/// A wrapper around `Storage` that tracks associated SPIR-V <id>
+/// with the elements, allowing the lookup of `Token<T>` by <id>.
+pub struct LiftStorage<T, L = Token<T>> {
+    values: Storage<T>,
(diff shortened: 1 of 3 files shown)


rspirv/lift/storage.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::{
+    borrow::Borrow,
+    collections::hash_map::{Entry, VacantEntry, HashMap},
+    hash::BuildHasherDefault,
+};
+
+use fxhash::FxHasher;
+
+
+pub struct LiftStorage<T, L = Token<T>> {
```
antiagainst: Can we please have documentation for public APIs?

rspirv/lift/storage.rs
```diff
@@ -0,0 +1,59 @@
...
+    ) -> (Token<T>, VacantEntry<spirv::Word, L>) {
+        let token = self.values.append(value);
+        match self.lookup.entry(id) {
+            Entry::Occupied(_) => panic!("Id {:?} is already used", id),
+            Entry::Vacant(e) => (token, e),
+        }
+    }
+}
+
+impl<T> LiftStorage<T> {
```
antiagainst: It has been a while and this syntax is not quite familiar to me anymore. What's the difference of this `impl` compared to the above one?

rspirv/lift/storage.rs
kvark: yes, sure 😅

rspirv/lift/storage.rs
kvark: The one above is generic across `L`. This block omits `L` thus it's for the default `L = Token<T>` only, which is a narrower subset of all.

PR #106, opened 2019-09-24
Lift operations, types, and constants

kvark:
Based on #105, but really quite far ahead.

Note: commits here are progressive and make sense to keep as is and not squash. As I was implementing them, I ensured `cargo test` at each step. We can think of another way to prove this.

### Summary

The core of this PR is auto-generating `lift_op` and `lift_type` methods that convert DR `Op` and `Type` into SR. The `lift_constant` method is also there but implemented manually. Lifting these 3 kinds of objects would constitute the "meat" of a SPIR-V module.

What is missing now is proper linking between them (i.e. all the use of `Token<>`) as well as lifting things like global variables.

### DIFF
--- a/autogen/sr.rs
+++ b/autogen/sr.rs
@@ -42,7 +42,7 @@ impl OperandTokens {
                     ),
                     //TODO:
                     //"Condition" => Token<Instruction>,
-                    //"Default" => Token<BasicBlock>,
+                    //"Default" => Token<Block>,
                     //"Value" => Token<Instruction>,
                     name if name.ends_with(" Label") => (
                         quote! { spirv::Word },
--- a/autogen/sr.rs
+++ b/autogen/sr.rs
@@ -31,6 +31,14 @@ impl OperandTokens {
                         quote! { Token<Constant> },
                         quote! { self.constants[value] },
                     ),
+                    "Field Types" => (
+                        quote! { StructMember },
+                        quote! { types::StructMember::new(value.clone()) },
+                    ),
+                    "Parameter Types" => (
+                        quote! { Token<Type> },
+                        quote! { *self.types.lookup(*value).1 },
+                    ),
                     // Function type is manually linked by the code.
                     "Function Type" => (
                         quote! { spirv::Word },
@@ -38,7 +46,7 @@ impl OperandTokens {
                     ),
                     name if name.ends_with(" Type") => (
                         quote! { Token<Type> },
-                        quote! { self.types[value] },
+                        quote! { *self.types.lookup(*value).1 },
                     ),
                     //TODO:
                     //"Condition" => Token<Instruction>,
@@ -254,6 +262,7 @@ pub fn gen_sr_code_from_instruction_grammar(
 
     let mut inst_structs = Vec::new();
     let mut op_variants = Vec::new();
+    let mut op_lifts = Vec::new();
     let mut branch_variants = Vec::new();
     let mut branch_lifts = Vec::new();
     let mut terminator_variants = Vec::new();
@@ -392,13 +401,23 @@ pub fn gen_sr_code_from_instruction_grammar(
 
             }
             _ => {
-                op_variants.push(if field_names.is_empty() {
-                    quote!{ #name_ident }
+                if field_names.is_empty() {
+                    op_variants.push(quote!{ #name_ident });
+                    op_lifts.push(quote! {
+                        #opcode => Ok(ops::Op::#name_ident),
+                    });
                 } else {
-                    quote! { #name_ident {
-                        #( #field_names: #field_types ),*
-                    }}
-                });
+                    op_variants.push(quote! {
+                        #name_ident {
+                            #( #field_names: #field_types ),*
+                        }
+                    });
+                    op_lifts.push(quote! {
+                        #opcode => Ok(ops::Op::#name_ident {
+                            #( #field_names: #field_lifts, )*
+                        }),
+                    });
+                }
             }
         }
     }
@@ -452,6 +471,15 @@ pub fn gen_sr_code_from_instruction_grammar(
                     #( #terminator_lifts )*
                     _ => self.lift_branch(raw)
                         .map(ops::Terminator::Branch)
+                }
+            }
+            pub fn lift_op(
+                &mut self, raw: &dr::Instruction
+            ) -> Result<ops::Op, InstructionError> {
+                let mut #iter_ident = raw.operands.iter();
+                match raw.class.opcode as u32 {
+                    #( #op_lifts )*
+                    _ => Err(InstructionError::WrongOpcode),
                 }
             }
             #( #lifts )*
--- a/autogen/sr.rs
+++ b/autogen/sr.rs
@@ -413,8 +413,8 @@ pub fn gen_sr_code_from_instruction_grammar(
                     });
                 }
             }
+            // Skip OpPhi as explicitly processed
             _ if inst_name == "Phi" => {
-
             }
             _ => {
                 if field_names.is_empty() {
--- a/autogen/sr.rs
+++ b/autogen/sr.rs
@@ -29,15 +29,15 @@ impl OperandTokens {
                 let (ty, value) = match operand.name.trim_matches('\'') {
                     "Length" => (
                         quote! { Token<Constant> },
-                        quote! { self.constants[value] },
+                        quote! { self.constants.lookup_token(*value) },
                     ),
                     "Field Types" => (
                         quote! { StructMember },
-                        quote! { types::StructMember::new(value.clone()) },
+                        quote! { StructMember::new(self.types.lookup_token(*value)) },
                     ),
                     "Parameter Types" => (
                         quote! { Token<Type> },
-                        quote! { *self.types.lookup(*value).1 },
+                        quote! { self.types.lookup_token(*value) },
                     ),
                     // Function type is manually linked by the code.
                     "Function Type" => (
@@ -46,7 +46,7 @@ impl OperandTokens {
                     ),
                     name if name.ends_with(" Type") => (
                         quote! { Token<Type> },
-                        quote! { *self.types.lookup(*value).1 },
+                        quote! { self.types.lookup_token(*value) },
                     ),
                     //TODO:
                     //"Condition" => Token<Instruction>,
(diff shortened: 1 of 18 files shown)


autogen/sr.rs
```diff
@@ -374,6 +374,9 @@ pub fn gen_sr_code_from_instruction_grammar(
                         }),
                     });
                 }
+            }
+            _ if inst_name == "Phi" => {
+
```
Skepfyr: I'm assuming this is a purposeful "ignore Phi instructions", if it is, a comment briefly explaining that (and maybe where they do get handled) would be a good idea.

rspirv/sr/autogen_ops.rs
```diff
@@ -768,9 +768,6 @@ pub enum Op {
         semantics: spirv::Word,
         value: spirv::Word,
     },
-    Phi {
```
Skepfyr: I'm beginning to feel that our approach to autogen is flawed, considering how many changes need to be made beyond the base information provided in the SPIR-V json.

rspirv/sr/module.rs
```diff
@@ -18,8 +18,18 @@ pub struct EntryPoint {
 
 #[derive(Debug)]
 pub struct BasicBlock {
-   pub ops: Vec<Op>,
-   pub terminator: ops::Terminator,
+    pub arguments: Vec<Token<Type>>,
```
Skepfyr: I really like this approach to basic blocks, it's lovely and get's rid of a load of oddities surrounding phi instructions.

rspirv/lift/mod.rs
```diff
@@ -24,11 +24,18 @@ use spirv;
 /// Reverse lookup table that associates SPIR-V <id> with SR tokens.
 type LookupMap<T> = HashMap<spirv::Word, Token<T>, BuildHasherDefault<FxHasher>>;
 
+struct OpInfo {
```
Skepfyr: This may be clarified in a later commit but it isn't at all obvious what this is doing.

rspirv/lift/storage.rs
```diff
@@ -0,0 +1,59 @@
+use crate::{
+    sr::{
+        storage::{Storage, Token},
+    },
+};
```
Skepfyr: I can't decide if that's beautiful or ugly.....

rspirv/sr/module.rs
```diff
@@ -39,7 +39,7 @@ pub struct Function {
     /// Function parameters.
     pub parameters: Vec<Token<Type>>,
     /// Basic blocks in this function.
-    pub basic_blocks: Vec<BasicBlock>,
+    pub basic_blocks: Vec<Token<BasicBlock>>,
```
Skepfyr: I'd be tempted to just store the first and let users traverse it as the graph that it is, but it doesn't matter too much.

rspirv/sr/module.rs
```diff
@@ -61,8 +61,11 @@ pub struct Module {
     /// All constants.
     pub constants: Storage<Constant>,
 
-    /// All functions.
-    pub functions: Vec<Function>,
+    /// All basic blocks.
+    pub basic_blocks: Storage<BasicBlock>,
```
Skepfyr: What is the reasoning behind having shared block storage between functions rather than per function? Having it per function would allow each function to be treated more separately, and moving blocks between pre-existing functions isn't exactly going to be a common operation.

autogen/sr.rs
```diff
@@ -299,14 +306,23 @@ pub fn gen_sr_code_from_instruction_grammar(
...
+                        #type_ident {
+                            #( #field_names: #field_types ),*
+                        }
+                    });
+                    type_lifts.push(quote! {
+                        #opcode => Ok(Type::#type_ident {
+                            #( #field_names: #field_lifts, )*
+                        }),
+                    });
+                }
```
Skepfyr: Hmmm, I've seen this exact code before... :grin: It shouldn't be abstracted yet but it's something to keep an eye on.

rspirv/lift/storage.rs
```diff
@@ -56,4 +56,10 @@ impl<T> LiftStorage<T> {
         entry.insert(token);
         token
     }
+
+    pub(in crate::lift) fn lookup_token(
```
Skepfyr: What's the value of this over normal indexing? Also, this looks like it will panic on invalid code, which I'm not sure about.

rspirv/lift/mod.rs
```diff
@@ -93,23 +132,60 @@ impl LiftContext {
...
+                let mut arguments = Vec::new();
+                for inst in &block.instructions {
+                    match inst.class.opcode {
+                        spirv::Op::Line => {} // skip line decorations
+                        spirv::Op::Phi => {
+                            match inst.operands[0] {
+                                dr::Operand::IdRef(id) => {
+                                    let (_, info) = context.ops.lookup(id);
+                                    arguments.push(info.ty
+                                        .expect("No return type for Phi source?")
```
Skepfyr: A question mark in an error message is a bit odd, also is this panicking if there's an error in the SPIR-V? That seems like a dangerous road to go down.

rspirv/sr/module.rs
Skepfyr: Although thinking about this more isn't this going to make instructions that use these much more complicated as they can't just refer to a previous operation? We may want to introduce an SSA Value type, it would also get ridd of all the weirdness of having to store instructions in Storages.

autogen/sr.rs
kvark: yes, absolutely!

autogen/sr.rs
kvark: will do

rspirv/lift/mod.rs
kvark: will do

rspirv/lift/mod.rs
kvark: good point, I'm fixing this

rspirv/lift/storage.rs
kvark: making this a little better now

rspirv/lift/storage.rs
kvark: I think panicking is the behavior today for any lookup in the storage. It's not great, but we should address this separately.

Just indexing is not possible because `lookup` field is private. The idea is that `values` and `lookup` should always be in sync here, so we don't expose either directly.

rspirv/sr/autogen_ops.rs
kvark: We had this discussion earlier in the SR path with @antiagainst , and it wasn't clear if it was worth it to just hand-write all the instructions manually. We decided to proceed with autogen modifications until it's more clear. So let's indeed keep an eye on the amount of exceptions and custom handling in the autogen.

rspirv/sr/module.rs
kvark: I'm not sure what you mean by "refer to a previous operation", please clarify?

As for SSA Value, please file an issue and describe what you have in mind.

kvark:
@Skepfyr I assume you are happy with the changes? Launched the travis runs on each commit, so that we don't have to squash them.
I think we'll still wait for @antiagainst to look at it on the Weekend or next week... hopefully, not too long

rspirv/sr/module.rs
kvark: Good point. I think that would make sense!

rspirv/sr/module.rs
kvark: I think you are correct. These should be moved to a function.

rspirv/sr/module.rs
Skepfyr: Currently, operations use `Token<Op>` to refer to the results of previous instructions? With this change, it's impossible for an instruction to refer to the result of a phi instruction as it isn't stored in the `Storage<Op>` any more. The simplest solution would be to just add some dummy ops to the storage.

kvark:
@antiagainst please review - everything is rebased now, and each commit is individually CI tested.

rspirv/sr/module.rs
kvark: Correct. We'll still have some `Op::Phi` things in the storage.

kvark:
@antiagainst I extended the summary as we discussed. There isn't much... it's hardly controversial either, I think. Please have a high-level look.

rspirv/lift/autogen_context.rs
```diff
@@ -95,6 +95,5492 @@ impl LiftContext {
...
+            148u32 => Ok(ops::Op::Dot {
+                vector_1: (match operands.next() {
+                    Some(&dr::Operand::IdRef(ref value)) => Some(*value),
+                    Some(_) => Err(OperandError::WrongType)?,
+                    None => None,
+                })
+                .ok_or(OperandError::Missing)?,
+                vector_2: (match operands.next() {
+                    Some(&dr::Operand::IdRef(ref value)) => Some(*value),
+                    Some(_)
```
Skepfyr: Lots of these are actually booleans in disguise, I think it would be better if SR exposed them as such but I'm not sure how best to add all that.

rspirv/sr/module.rs
antiagainst: +1 to what @Skepfyr said. I've also mentioned that we don't have a proper SSA value data type in structured representation at various places. (Also because the structured representation is under development right now so it's broken anyway. ;-P) Depending on how we want to model function parameters and Phi nodes (they are both instructions in SPIR-V spec and DR), we may end up having two more cases an instruction can reference as an operand. So having a value type, which is an enum among FunctionArgument, BlockArgument, and OpResult. And I can see this is how the current impl is heading for. (We baked function argument into `sr::Function` and block argument to `sr::Block`.) Another thing we need to think about is how to support def-use chain. Normally the use chain is associated with the value.

antiagainst:
Overall looks good to me! I'll leave the final approval to @Skepfyr. :) Thanks @kvark for all of this and sorry for the delay!

kvark:
@Skepfyr do you think we can merge and then proceed with adding the SSA value type? Or would you want to see that experimentation in the PR first? It feels like a major piece of work that isn't directly related.

Skepfyr:
I'm happy for this to be merged as is, the questions need addressing but I'm happy for them to be sorted out as we go along.

kvark:
This is rebased, every commit is checked for being testable independently.
Don't want this work to hang in the air, proceeding.

Issue #108, opened 2019-09-27
Publish 1.3.5 of spirv-headers


Jasper-Bekkers:
Now that #97 and #100 landed I would love to be able to depend on spiv-headers proper again. Would you mind publishing a new version?

kvark:
I don't have a problem with that. Changes seem backwards compatible. @antiagainst anything to check before we proceed?

antiagainst:
SGTM! Please feel free to proceed. The version number is tracking the spec version.  So the next one would be 1.4.1. One thing to keep in mind is that we use a custom verison of [core grammar](https://github.com/gfx-rs/rspirv/blob/master/autogen/external/spirv.core.grammar.json) (with categories added and adjusted a bunch of names there) so bumping the version will require manually merge upstream changes with our version. (I'm sorry about this. The changes should really be upstreamed, at least those components that make sense.) I'd recommend going into 1.4.1 instead of 1.5.1 (which is the current spec version) so that we can minimize the changes needed to merge and also a consumer can specifically pin to this version if needed. The procedure would be: diff the current in-tree grammar with the current tracked upstream grammar, update the git submodule for spirv-headers to https://github.com/KhronosGroup/SPIRV-Headers/commit/c4f8f65792d4bf2657ca751904c511bbcf2ac77b and then apply the changes to get a newer version of the custom grammar.

Jasper-Bekkers:
Wouldn't it make sense to publish both 1.4.1 and 1.5.1 so we're up to date with the latest spec version as well?

antiagainst:
Certainly. My previous point was that let's not omit 1.4.1 and jump to 1.5.1 directly. :) We should update to 1.5.1 to catch up with the spec for sure. :)

Jasper-Bekkers:
I've submitted 2 PRs #115 and #116 that update the json files.

azriel91:
Heya, please may I bump this :wave:. The current `spirv_headers` version on crates.io depends on `num-derive 0.1.44` which brings in `syn 0.12.15`. I'm slowly chewing through updating the Rust ecosystem :smile:

kvark:
@azriel91 so you are blocked on #116?

azriel91:
Heya, yeaps.

antiagainst:
The `syn` dependency is due to `derive_more` used by `rspirv`. If one bumps the version of `derive_more` to `0.99` then `syn` will go away. But that results in some compilation failures I haven't gotten time to look into it.

[...]

PR #115, opened 2019-10-21
Update SPIR-V to 1.4.1

Jasper-Bekkers:
Some legwork for #108 somebody will still need to fix the submodule to point to https://github.com/KhronosGroup/SPIRV-Headers/commit/c4f8f65792d4bf2657ca751904c511bbcf2ac77b

### DIFF
--- a/autogen/external/spirv.core.grammar.json
+++ b/autogen/external/spirv.core.grammar.json
@@ -569,7 +569,6 @@
       "operands" : [
         { "kind" : "IdRef",                            "name" : "'Target'" },
         { "kind" : "IdRef",                            "name" : "'Source'" },
-        { "kind" : "MemoryAccess", "quantifier" : "?" },
         { "kind" : "MemoryAccess", "quantifier" : "?" }
       ]
     },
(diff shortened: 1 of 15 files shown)


autogen/external/spirv.core.grammar.json
```diff
@@ -132,7 +132,7 @@
         { "kind" : "IdResult" },
         { "kind" : "IdRef",                                     "name" : "'Set'" },
         { "kind" : "LiteralExtInstInteger",                     "name" : "'Instruction'" },
-        { "kind" : "IdRef",                 "quantifier" : "*", "name" : "Operands" }
+        { "kind" : "IdRef",                 "quantifier" : "*", "name" : "'Operand 1', +\n'Operand 2', +\n..." }
```
kvark: is this really the name of an operand? We have some code that tries to make decisions based on that field.

Jasper-Bekkers: That's what's in upstream, so I suggest we patch up that code to make it easier to merge in the future.

Jasper-Bekkers: If it's a massive complication I can ping  Khronos internally and see if we can change it back.

kvark: I think we can get around by checking the `Op` itself instead of the argument name

Jasper-Bekkers:
I think that's be the cleaner option then

Jasper-Bekkers:
I think I finished up the rest of the upgrade; I'm not sure if anything else needs to be done but I'd be willing to put in some more time if needed.

Jasper-Bekkers:
It might make sense to add @antiagainst as a reviewer as well in this case.

autogen/external/spirv.core.grammar.json
antiagainst: Yes. That field is used to generate the spec. So I'm not sure we can convince upstream to change it.

PR #116, opened 2019-10-21
Merge SPIR-V json to 1.5.1

Jasper-Bekkers:
This is based on my PR #115 and it's a pretty major one since they added "class" attributes to a lot of elements and some are named differently then what they were in rspirv, so I chose the names from upstream instead to make future divergence easier to manage.

Submodules should be updated to https://github.com/KhronosGroup/SPIRV-Headers/commit/63d4d272f6e5b3cb9bb2bb50718a886a3eef4dab

This should help with issue #108

### DIFF
--- a/autogen/external/spirv.core.grammar.json
+++ b/autogen/external/spirv.core.grammar.json
@@ -26,16 +26,122 @@
   ],
   "magic_number" : "0x07230203",
   "major_version" : 1,
-  "minor_version" : 4,
+  "minor_version" : 5,
   "revision" : 1,
+  "instruction_printing_class" : [
+    {
+      "tag"     : "@exclude"
+    },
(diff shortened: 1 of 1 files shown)


antiagainst:
I've pushed out v.1.4.2 so this now has conflicts. Yeah, v1.5 is a substantial change to the grammar file. I think it's the time for us to ditch our own local copy of spirv.core.grammar.json now. Do you have time to look into this?

antiagainst:
I mean switching to use upstream spirv.core.grammar.json?

Jasper-Bekkers:
Unfortunately I won't have time for this in the foreseeable future :-(

Jasper-Bekkers:
This is superseded by #132

PR #118, opened 2020-02-14
Lift more things

kvark:


rspirv/lift/mod.rs
```diff
@@ -131,11 +131,15 @@ impl LiftContext {
...
                     .ok_or(ConversionError::MissingFunction)?
             )?;
-            //TODO: lift function type instruction
 
             for block in fun.blocks.iter() {
                 let mut arguments = Vec::new();
-                for inst in &block.instructions {
+                let mut ops = Vec::new();
+                if block.instructions.is_empty() {
+                    return Err(ConversionError::MissingTerminator);
```
Skepfyr: I'm worried this won't get caught if the terminator gets moved out of instructions. It will suddenly start erroring on empty blocks.

rspirv/lift/mod.rs
```diff
@@ -131,11 +131,15 @@ impl LiftContext {
                     .as_ref()
                     .ok_or(ConversionError::MissingFunction)?
             )?;
-            //TODO: lift function type instruction
```
Skepfyr: I can't see the change lifting the function type instruction?

PR #119, opened 2020-02-14
SR names

kvark:
Based on #118


Jasper-Bekkers:
@kvark is this still something you're interested in merging?

Issue #123, opened 2020-03-16
Derive serialization for the types in spirv-headers


kvark:
Would be great to have an optional `serde` feature, which makes all the types from headers to derive serialization.

paulkernfeld:
I think this is now implemented by #125 👍

Issue #21, opened 2017-08-03
Name of the crate


neon64:
First off, this isn't exactly an issue per se.

I was looking through my crates on `crates.io` and discovered I had a crate from 2 years ago named `spirv` - which I created just to support a hobby project before there were any other SPIR-V libraries in Rust. I haven't maintained it in ages and the code/features/testing/etc are quite obviously inferior to the fantastic work being done in this crate. So basically my question is, since I'm going to mark that as 'abandoned' and suggest that nobody should use it, would another crate (eg: this one) like to take that `spirv` name?

I realise this is perhaps a dumb question, since changing names on crates.io would force all the users of `rspirv` to change their code, however I guess I'm just letting you know in case somebody feels that `spirv` is a 'nicer' or more canonical name to use.

*Edit:* I guess I'm also looking for https://github.com/rust-lang/rust/issues/41616

antiagainst:
Hi @neon64, much appreciated your generosity and kindness to transfer the `spirv` crate name! `spirv` sounds like a nicer name for the [header sub-crate](https://crates.io/crates/spirv_headers) in this repo. However, as you said, it is rather intrusive to change crate names under the current cargo/crate management settings. So I'm just not sure it worth the hassle.

neon64:
Okay no worries.. Well maybe for now I'll just put a clear link to this project suggesting people go use it instead.

antiagainst:
Thanks @neon64! :)

antiagainst:
Hey @neon64, are you still willing to shift the ownership of the `spirv` crate to us now? Previously there isn't a need, but now it looks the need does arise: `spirv_headers` is using the SPIR-V version as the crate version; it has seen a few breakages thus far due to the differences from semver. It might be better to start with a clean slate to push to `spirv` crate where we can adopt semver from now on. WDYT?

neon64:
@antiagainst I'd be fine to transfer ownership. Which username(s) should I invite?

Jasper-Bekkers:
@Jasper-Bekkers, @kvark and @antiagainst at least I think :-)

antiagainst:
Yes. Thanks @neon64!

PR #125, opened 2020-03-26
Derive serialization for the types in spirv-headers

Nanopro:
Based on #123 

Added serde as a optional dependency in spirv-header,
and feature gated derive for all enums and structs.

### DIFF
--- a/spirv/Cargo.toml
+++ b/spirv/Cargo.toml
@@ -11,6 +11,10 @@ readme = "README.md"
 license = "Apache-2.0"
 keywords = ["spirv", "definition", "struct", "enum"]
 
+[features]
+serialize = ['serde']
+deserialize = ['serde']
+
 [lib]
 path = "lib.rs"
 
--- a/spirv/Cargo.toml
+++ b/spirv/Cargo.toml
@@ -17,3 +17,4 @@ path = "lib.rs"
 [dependencies]
 bitflags = "1"
 num-traits = "0.2"
+serde = {version = "1", optional = true, features = ["derive"]}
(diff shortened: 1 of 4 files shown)


kvark:
@antiagainst could you take a look please?

Nanopro:
@kvark Is it good now?

kvark:
Yes, that's more future-proof, thanks! Please make sure the path is CI tested.
Also, I really want @antiagainst to look at this before merging.

Nanopro:
@kvark I am not sure that error in CI is caused by my commits.

kvark:
indeed! filed #126

spirv/Cargo.toml
```diff
@@ -11,9 +11,14 @@ readme = "README.md"
...
+serialize = ['serde']
+deserialize = ['serde']
+
 [lib]
 path = "lib.rs"
 
 [dependencies]
 bitflags = "1"
 num-traits = "0.2"
+serde = {version = "1", optional = true, features = ["derive"]}
```
kvark, line 24: note that this isn't correct, feature should be `serde_derive`, or the version needs to be more specific

kvark:
@antiagainst how do we get this out for users?

antiagainst:
@kvark: I see we might want more breaking changes along the way so it's better to sort out the versioning issue. I'm thinking maybe it's better to retire the current spirv_header crate and instead publish to spirv crate so that we have a chance to detach the spec version and the crate version. @neon64 asked me previously in #21 but back then we don't have a clear need to make it worthwhile. Now maybe its the time.

kvark:
I haven't given it a lot of thought, tbh. It sounded to me, based on the previous discussions, that relying on upstream Spec semantic versioning can be a problem. In this case, yes, we have to detach it. Do you own "spirv" crate now? Sounds like a good plan :)

PR #127, opened 2020-04-28
Install and run rustfmt only on stable Rust

### LINKED ISSUE #126: Fix CI failure on nightly with rustfmt
https://travis-ci.org/github/gfx-rs/rspirv/jobs/667493315?utm_medium=notification&utm_source=github_status

Looks like we need to have Nightly failable, or better - just not try to use rustfmt on nightly target.

paulkernfeld:
The goal of this change is to fix #126

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ jobs:
       rust: nightly 
       script: # rustfmt is not always available on nightly
         - cargo check --verbose
-        - git diff --exit-code
         - cargo test --verbose
     - os: osx
       rust: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,36 +10,21 @@ before_script:
 
 jobs:
   include:
-    - stage: install_rustfmt
-      os: linux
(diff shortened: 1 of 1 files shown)


paulkernfeld:
I'm guessing that this won't pass on the first try...

paulkernfeld:
This isn't ready to be merged yet. I think the Travis CI build is passing because it's not correctly running all the build steps.

paulkernfeld:
Since each stage contains multiple jobs (I had wrongly thought that a job would contain multiple stages), trying to make one stage for installing rustfmt and one stage for running the tests results in [7 jobs instead of 4](https://travis-ci.org/github/gfx-rs/rspirv/builds/680679299), which I think is probably not desirable. I think it's possible to to do this while still having only 4 jobs. I'm going to investigate those possibilities.

paulkernfeld:
Okay, I think the most recent commit should fix the issue. 😀 Please let me know if I can make any additional changes.

PR #132, opened 2020-06-18
Update to upstream SPIR-V 1.5.2

Jasper-Bekkers:
This removes the internal version of spirv.core.grammar.json and switches to use only the upstream one spirv.core.grammar.json file as discussed in #116 as a potential better way forward.

This also updates all the bindings to SPIR-V 1.5.2

PR #133, opened 2020-06-19
Allow Builder to modify modules

Jasper-Bekkers:
This sits on top of #132 

This would allow the `Builder` to also be used to modify existing modules, making it useful for, for example, runtime patching of SPIR-V code.

The biggest additions so far are: 
 - One can create a Builder from an existing Module
 - Relative insert of instructions; besides the regular `append` style instructions there is now an extra set of instruction builders that take a relative offset
 - There's a (very simple) mechanism to select basic blocks and functions for editing

Code sample of the new APIs:

```
builder.insert_load(InsertPoint::FromEnd(2), int32_id, Some(previously_allocated_result_id), load_location, None, &[]).unwrap();
```

This would insert a `OpLoad` as the second to last instruction (for example, right before an OpReturn that may already be there).

Existing builder functionality / API stays in tact; so `builder.load(...)` would still append a load to basic block currently under construction.


Jasper-Bekkers:
If you're reviewing this is probably a more useful view since it skips the changes for #132 https://github.com/Traverse-Research/rspirv/compare/master...Traverse-Research:builder-mods

autogen/dr.rs
```diff
@@ -71,6 +71,20 @@ fn get_function_name(opname: &str) -> TokenStream {
     }
 }
 
+/// Returns a suitable function name for the given `opname`.
+fn get_function_name_with_prepend(prepend: &str, opname: &str) -> TokenStream {
+    if opname == "OpReturn" {
```
kvark, line 76: nit: this cold be `let name = match opname {...};`

rspirv/dr/build/autogen_type.rs
```diff
@@ -3,86 +3,130 @@
 // DO NOT MODIFY!
 
 impl Builder {
-    #[doc = "Appends an OpTypeVoid instruction and returns the result id."]
+    #[doc = "Appends an OpTypeVoid instruction and returns the result id, or return the existing id if the instruction was already present."]
     pub fn type_void(&mut self) -> spirv::Word {
+        let mut inst = dr::Instruction::new(spirv::Op::TypeVoid, None, None, vec![]);
+        for ty in &self.module.types_global_values {
```
kvark, line 9: This PR essentially introduces de-duplication logic to DR. Is this desired? It adds a linear search over existing types, which makes the run-time complexity rather unfortunate.
Given that DR's goal is to originally be able to quickly construct a SPIR-V module, this de-duplication seems to be higher level and not really necessary. Something like an optimization pass (spirv-opt?) could take care of this automatically, and probably also faster than running this code on each new type.

rspirv/dr/build/autogen_type.rs
```diff
@@ -3,86 +3,130 @@
...
             None,
-            Some(id),
+            None,
             vec![
                 dr::Operand::LiteralInt32(width),
                 dr::Operand::LiteralInt32(signedness),
             ],
-        ));
+        );
+        for ty in &self.module.types_global_values {
```
kvark, line 47: if we are to add de-duplication (which sounds iffy to me), we'd at least want that code shared, i.e. have a command function to add a type instruction that we'd call from these, and which would try to deduplicate

rspirv/dr/build/mod.rs
```diff
@@ -90,23 +90,92 @@ type BuildResult<T> = result::Result<T, Error>;
...
+
+        let ref mut block = if self.new_block.is_some() {
+            self.new_block.as_mut().unwrap()
+        } else {
+            &mut self.module.functions[self.selected_function.unwrap()].blocks[self.selected_block.unwrap()]
+        };
+
+        match insert_point {
+            InsertPoint::End => block.instructions.push(inst),
+            InsertPoint::Begin => block.instructions.insert(0, inst),
```
kvark, line 152: This is also concerning: an application that adds N instructions to a block of M will spend O(N*M) time moving bytes around. I don't think this is acceptable to DR level of abstraction.

rspirv/dr/build/autogen_type.rs
Jasper-Bekkers: I'm not sure if spirv-opt can do it, because it's invalid to have spirv with duplicate non-aggregate types.

https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#TypesVariables

I added this because otherwise this needs to end up in userland for everybody building SPIR-V modules, and because it's a pretty small search typically.

rspirv/dr/build/autogen_type.rs
Jasper-Bekkers: I can make that happen :-)

rspirv/dr/build/autogen_type.rs
Jasper-Bekkers: One option would be to add `type_void_dedup` helpers.

kvark:
I didn't realize the duplicate types were disallowed by the spec... That means we need to either:
  1. force this upon the user to deduplicate and assume uniqueness
  2. look for a duplicate on every insertion (by linear search or something like a hash map)
  3. do a pass at the time when building finishes, de-duplicating all the types

rspirv/dr/build/mod.rs
Jasper-Bekkers: Since builder is the most ergonomic way of building / patching SPIR-V that we have this kind of functionality is needed. One other way that's kind of iffy is to be able to insert bb's in arbitrary places, or just change the underlaying datatype to something more amenable to inserts in arbitrary places. I didn't change the datatype because I'd rather have the "new" code be slow then to perf-break existing code in this case.

Jasper-Bekkers:
> 
> 
> I didn't realize the duplicate types were disallowed by the spec... That means we need to either:
> 
>     1. force this upon the user to deduplicate and assume uniqueness
> 
>     2. look for a duplicate on every insertion (by linear search or something like a hash map)
> 

Yeah spirv-val called me out on this - so right now I only do the search for simple non-aggregate types, which you're most likely to just have set up in the beginning when you're building types.

>     3. do a pass at the time when building finishes, de-duplicating all the types

This seems undesirable - it would mean having to through and patch up all types referencing this.

rspirv/dr/build/mod.rs
kvark: I see, that makes sense.
Would it be good to record the list of inserted instructions on the side, so that we can merge them in a single pass at the end of building, instead of moving bytes around on every insertion?

Jasper-Bekkers: I think having a delayed insertion scheme will be much more bug prone; it's probably easier to replace the Vec with a VecDeque and comment that the relative insertion points incur a performance overhead.

kvark: I guess we can follow up with this without braking the API, so it's not a blocker anyway

Jasper-Bekkers:
Just squash and merge then, or what's the plan? :-)

kvark:
Yeah, that's fine by me. Just wanted to make sure that
  1. it's fine with you as well. Sometimes separate commits still make more sense
  2. you don't accidentally push/merge something that doesn't respect that workflow, now or in the future

Jasper-Bekkers:
Nah - I'm fine with squashing stuff. Sometimes it makes sense, but good to know for the future :-)

PR #134, opened 2020-07-03
dr: Re-add missing id in builder_types

MarijnS95:
The assignment of `id` for non-aggregate types is missing, resulting in build errors when compiling `autogen_type.rs`.

The desired `id` to return is not `self.id()`, but the result of `dedup_insert_type` which returns an existing `id` or `self.id()` if not found.

### DIFF
--- a/autogen/dr.rs
+++ b/autogen/dr.rs
@@ -330,7 +330,8 @@ pub fn gen_dr_builder_terminator(grammar: &structs::Grammar) -> TokenStream {
     let kinds = &grammar.operand_kinds;
     // Generate build methods for all types.
     let elements = grammar.instructions.iter().filter(|inst| {
-        inst.class == Some(structs::Class::Terminator) || inst.class == Some(structs::Class::Branch)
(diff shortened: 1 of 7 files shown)


kvark:
Why is this not caught by CI?

Jasper-Bekkers:
@kvark Looks like there are a few more issues with CI, we'll take a look at this on monday and checkout the CI setup for this project as well to see if it can be improved.

kvark:
Ah ok, that sounds familiar. We've had this issue with travis since they started moving to travis.com, and this is how they force everybody to upgrade or move to Github actions (and that's what I did for my projects, not looking back).
When I was merging/looking at last rspirv PRs, I was going to Travis panel and checking the status manually. The signalling is broken.

MarijnS95:
Pushed the necessary changes to make the test compile and succeed. Followup PRs include a CI change to GitHub actions and an autoformat. `cargo check` is covered (as in the Travis configuration), `cargo clippy` is not.

PR #135, opened 2020-07-07
Migrate CI to GitHub Actions

MarijnS95:
Depends on #134

Change the CI from Travis to GH Actions as discussed in #134, because Travis stopped signalling results to PRs. This fixes a minor nit and some warnings from `cargo check` as well. `cargo clippy` is not covered, the autogenerated files introduce far too many warnings. Perhaps we should exclude clippy from checking these files?

`cargo fmt` support will be introduced in a separate PR, as the diff has gotten quite large.


MarijnS95:
As far as I can tell these actions only start running when the configuration is merged to the target repository, that's why nothing shows up just yet :disappointed:

PR #136, opened 2020-07-07
Cargo fmt all files and enforce this in the CI

MarijnS95:
Not sure if we want to merge this directly, @kvark has three open PRs and I don't want to conflict with that unless you're okay updating those.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,5 +25,9 @@ jobs:
           args: --workspace --all-targets --verbose
       - uses: actions-rs/cargo@v1
         with:
+          command: fmt
+          args: --all -- --check
+      - uses: actions-rs/cargo@v1
+        with:
           command: test
           args: --workspace --all-targets --verbose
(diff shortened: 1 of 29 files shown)


kvark:
I don't have plans on those PRs, since we are no longer invested in rspirv DR/MR layers.
If they can land, ok. If they can't that's fine by me too.
You can check out their contents and figure out if the code is in line with what you have in mind.

MarijnS95:
@kvark Frankly speaking I haven't invested much time in using `rspirv` nor diving into the SPIR-V specification just yet, but may be able to comment on that in the future. As it stands two of the PRs have merge conflicts, doing formatting changes might make that an even larger barrier.
For now I'm only interested in formatting the entire repo to make future development easier - I accidentally created a large diff many times by hitting the autoformatting key while working on the changes for #134.

Perhaps @Jasper-Bekkers has an opinion on this matter, too.

Jasper-Bekkers:
@kvark if they are stale maybe we should just close them?

kvark:
My question is basically whether you see any value in them. If yes, I can try updating them for landing.
Fine with closing otherwise.

PR #139, opened 2020-07-23
Pass in which instruction got detached

Jasper-Bekkers:
This improves the DetachedInstruction error message because we occasionally run into it.

Also adds an ignore for OpLine since it wasn't handled properly before.

### DIFF
--- a/rspirv/dr/loader.rs
+++ b/rspirv/dr/loader.rs
@@ -17,7 +17,7 @@ pub enum Error {
     NestedBlock,
     UnclosedBlock,
     MismatchedTerminator,
-    DetachedInstruction,
+    DetachedInstruction(Option<dr::Instruction>),
     EmptyInstructionList,
     WrongOpCapabilityOperand,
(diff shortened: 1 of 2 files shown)


rspirv/dr/loader.rs
```diff
@@ -32,26 +32,32 @@ impl Error {
     ///
     /// This method is intended to be used by fmt::Display and error::Error to
     /// avoid duplication in implementation. So it's private.
-    fn describe(&self) -> &str {
-        match *self {
-            Error::NestedFunction => "found nested function",
-            Error::UnclosedFunction => "found unclosed function",
-            Error::MismatchedFunctionEnd => "found mismatched OpFunctionEnd",
+    fn describe(&self) -> String {
```
kvark, line 35: We could also return `Cow<'static, str>` here to avoid allocation of strings where not necessary

Jasper-Bekkers: Done

PR #142, opened 2020-08-03
GitHub: Add dependabot configuration for update notifications

MarijnS95:
Use dependabot to notify us about dependency updates.

Note that we might want to turn down the frequency to monthly and exclude update notifications for transitive (indirect) dependencies, before dependabot goes crazy on `Cargo.toml` and all the patch releases :grin:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  allow:
+  - dependency-type: all
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.10"
+version = "0.7.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -13,7 +13,7 @@ name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


MarijnS95:
We'll unfortunately be unable to update `derive_more` to `0.99` since they dropped automatic omitting of duplicate `impl From<>` in https://github.com/JelteF/derive_more/commit/6d43b1e1142eaebba2a0739f97cbd50ab64da44d in favour of manually specifying `#[from(ignore)]` on all duplicates. We'd have to replicate that functionality and emit the attribute ourselves, or ask whether this can be added back (behind an optional flag).

kvark:
What's the motivation for being notified about dependency patch updates?

MarijnS95:
@kvark Good question. We are running into trouble with some projects including older dependencies, and want to prevent that from happening by getting a little nudge when these can be updated. In particular type clashes between two versions of the same library, and having duplicate dependencies at all tanking compile times.

That said getting notifications for every patch release results in way too much noise considering that patch versions are not listed in `Cargo.toml` but `Cargo.lock` is checked in (which is ignored when linking against one of the libs in this package anyway). We should probably increase the interval drastically, though I can't find a way to limit dependabot to major version bumps (ie. those not automatically performed by `cargo update` following `Cargo.toml` version restrictions).

Given the limited dependencies in the project this might not be worth it... :thinking:

kvark:
I haven't worked on a project with a good demonstration of dependabot, and I'm struggling to see its usefulness for Rust projects. Let me try to explain, why.

If you have an application, and you want to always use the latest patch version of the up, having dependabot asking you to do this in *every* Cargo toml of *every* dependency of yours is a huge waste. Instead, you can simply do `cargo update` in a cron job or something on your side, and be done with it. And if we are talking about the breaking non-patch versions, dependabot can't provide a working patch anyway, it can at most notify you that a new version is out. Updating to every breaking release urgently is also a waste, so typically there is a big gap between a breaking version published and picked up by big applications. In this case, the benefit of quick notification is not obvious.

Note: `rspirv` includes Cargo lock for the binary target `rspirv-dis`, not for the libraries.

MarijnS95:
Let's drop this then, even if it would be beneficial (for us) to update `derive_more` which includes a whole chain of older packages, resulting in aforementioned duplicates and prolonged compile times.

kvark:
It's definitely good to track your duplicated dependencies. This is what would be driving the need to update `derive_more`. It's a better motivator than the hint from dependabot saying "hey, there is a new release of something".

PR #147, opened 2020-08-11
Properly deduplicate type declarations

khyperia:
https://github.com/gfx-rs/rspirv/pull/133/commits/716d74bed6fb5a0ec4ad66f6344f6c840bb0d48d broke deduplication, so, re-fix it. Also fix function deduplication by moving `#(#extras)*` *before* deduplication checks, not after. Also deduplicate aggregate types too. Also did a little micro-optimization to use `extend` instead of repeated `push` (helps with number of allocations).

### DIFF
--- a/autogen/dr.rs
+++ b/autogen/dr.rs
@@ -162,9 +162,7 @@ fn get_push_extras(
                     } else {
                         let kind = get_dr_operand_kind(&param.kind);
                         Some(quote! {
-                            for v in #name.as_ref() {
-                                #container.push(dr::Operand::#kind(*v));
-                            }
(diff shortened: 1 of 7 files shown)


kvark:
Sorry, I can't review all the rspirv changes, since as I mentioned, we aren't invested in it at all.
I'd be fine moving the repository under a different organization.

Jasper-Bekkers:
@kvark No worries, I can do it. I thought you may be interested in these, but other than that I'll do it.

Jasper-Bekkers:
@khyperia if we're going to support other (aggregate) types, we probably should update `is_type_identical` as well.

PR #148, opened 2020-08-12
Add OpFunctionCall explicitly

Jasper-Bekkers:
Most ops tagged with the Function class are models by our `Function` struct (eg. to create functions) however, OpFunctionCall also has the "Function" class, but should exist in the Builder.


Jasper-Bekkers:
@kvark since me and @khyperia will be working on rspirv quite a bit going forward, do you think we can add her as a maintainer to this repo as well?

kvark:
Do you maybe want it to be transferred under your org? I added @khyperia anyway.

Issue #143, opened 2020-08-07
Publish a new version of rspirv


Jasper-Bekkers:
Now that SPIR-V 1.5 has been in mainline for a while, we should probably publish a new version of this crate.

kvark:
Just make sure to triage the issues that could have impacted the API, to see if they need to be solved before publishing

Jasper-Bekkers:
I can take a look at #131 and  #121, but I think neither should be show stoppers for a release.

Jasper-Bekkers:
Do we need @antiagainst to push a new version or can you do it @kvark ?

antiagainst:
Thanks @Jasper-Bekkers and @kvark for taking of the issues! Unfortuantely I won't have the bandwidth to come back to this. I've invited @kvark on crates.io so he should be able to publish new versions now. :)

kvark:
@antiagainst thank you!
I'm fine publishing spirv_headers, but I'm not feeling positive about publishing `rspirv` itself, since we aren't using it anywhere. Perhaps, it would make sense to share the publishing joy of it with @Jasper-Bekkers ?

Jasper-Bekkers:
@kvark fwiw, we've been using latest on a few internal projects both at Traverse Research and Embark Studios now.

PR #149, opened 2020-08-14
Make ZeroOrMore take Iterator instead of AsRef<[T]>

khyperia:
This is useful if the user has an iterator instead of a slice/vec, it
saves an allocation.

I think this is a breaking change due to ref differences, not sure if we're okay with that.

(For review, the only file changed is `dr.rs`, the rest are generated)

### DIFF
--- a/autogen/dr.rs
+++ b/autogen/dr.rs
@@ -47,7 +47,7 @@ fn get_param_list(
                     structs::Quantifier::One => quote! { #name: #kind },
                     structs::Quantifier::ZeroOrOne => quote! { #name: Option<#kind> },
                     structs::Quantifier::ZeroOrMore => {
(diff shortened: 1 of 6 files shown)


Jasper-Bekkers:
I don't quite know what we rspirv typically does with breaking changes - nor do I know if the change from AsRef<[T]> to IntoIterator is a breaking change.

However, I'm inclined to accept it because this crate is already a bit on the alloc-heavy side of things, so getting that down would be quite nice.

kvark:
Looks like a breaking change to me, given that `AsRef<[T]>` would be implemented for `&[T]` but `IntoIterator<T>` wouldn't.

Jasper-Bekkers:
Might be good to merge this in after #143

PR #158, opened 2020-09-02
Add methods for iterating over all instructions in the module

Jasper-Bekkers:
Fixes incorrect doc comments left over from previous refactor in https://github.com/gfx-rs/rspirv/commit/a4d9cdd5a21a774051e30cba13da6612396859a1. Added a few more helper methods to iterate over / list all `Instruction`'s in a `Module.

### DIFF
--- a/rspirv/dr/constructs.rs
+++ b/rspirv/dr/constructs.rs
@@ -120,9 +120,6 @@ impl Module {
     }
 
     /// Returns an iterator over all global instructions.
-    ///
-    /// This method internally creates a vector of references to all global
-    /// instructions, therefore it has some overheads.
     pub fn global_inst_iter(&self) -> impl Iterator<Item = &Instruction> {
         self.capabilities
             .iter()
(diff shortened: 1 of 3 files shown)


rspirv/dr/constructs.rs
```diff
@@ -135,6 +132,70 @@ impl Module {
...
+            .chain(&self.annotations)
+            .chain(&self.types_global_values)
+            .chain(self.functions.iter().flat_map(|f| {
+                f.def
+                    .iter()
+                    .chain(f.parameters.iter())
+                    .chain(
+                        f.blocks
+                            .iter()
+                            .flat_map(|b| b.label.iter().chain(b.instructions.iter())),
```
khyperia, line 169: took me a bit to remember `Option` is `IntoIterator` for the label here :P

khyperia (approved):

rspirv/dr/constructs.rs
```diff
@@ -135,6 +132,70 @@ impl Module {
...
+        self.capabilities
+            .iter()
+            .chain(&self.extensions)
+            .chain(&self.ext_inst_imports)
+            .chain(&self.memory_model)
+            .chain(&self.entry_points)
+            .chain(&self.execution_modes)
+            .chain(&self.debugs)
+            .chain(&self.annotations)
+            .chain(&self.types_global_values)
```
MarijnS95, line 161:
```suggestion
        self.global_inst_iter()
```

Can't do this for `all_inst_iter_mut` though because `self` would be mutably borrowed for the duration of the result of `global_inst_iter_mut` and can't be mutably borrowed again for `self.functions.iter_mut()` :disappointed:

rspirv/dr/constructs.rs
Jasper-Bekkers: Yeah it's nice Option can do that, but it's not always intuitive.

rspirv/dr/constructs.rs
Jasper-Bekkers: I kept it duplicated because I think it's easier to keep them in sync that way - rather then having to have a special `mut` version.

rspirv/dr/constructs.rs
MarijnS95: Monads :tada:

rspirv/dr/constructs.rs
MarijnS95: Makes sense; I was hoping you had an idea/solution for the mutability problem. Makes me wonder why borrowing `self` mutable multiple times for every argument to `chain()` is fine, while invoking a function (that should merely propagate the same "mutable lifetime scope") is not.

MarijnS95: It makes sense now: The compiler understands that distinct members of `self` are borrowed in the `.chain()` case. When the function is called however, `self` in its entirety is borrowed mutably.
Duplicating one of the `.chain()` invocations results in a compiler error: that field is already mutably borrowed.

I don't think there's an easy solution to this, let's leave the duplication in place.

PR #161, opened 2020-09-02
Add support for 64-bit OpSwitch

khyperia:
SR support is omitted for now.

I'm not 100% the parser is fully working, I couldn't figure out how to write tests - but the emitter works (which is the easy side).

### DIFF
--- a/autogen/binary.rs
+++ b/autogen/binary.rs
@@ -238,11 +238,7 @@ pub fn gen_operand_parse_methods(grammar: &Vec<structs::OperandKind>) -> TokenSt
     // Logic operands that expand to concrete operand pairs,
     // that is, those operand kinds with 'Pair' name prefix.
     // We only have three cases. So hard code it.
-    let pair_kinds = vec![
-        ("LiteralInteger", "IdRef"),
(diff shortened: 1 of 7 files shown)


autogen/binary.rs
```diff
@@ -238,11 +238,7 @@ pub fn gen_operand_parse_methods(grammar: &Vec<structs::OperandKind>) -> TokenSt
     // Logic operands that expand to concrete operand pairs,
     // that is, those operand kinds with 'Pair' name prefix.
     // We only have three cases. So hard code it.
-    let pair_kinds = vec![
-        ("LiteralInteger", "IdRef"),
-        ("IdRef", "LiteralInteger"),
-        ("IdRef", "IdRef"),
-    ];
+    let pair_kinds = vec![("IdRef", "LiteralInteger"), ("IdRef", "IdRef")];
```
Jasper-Bekkers, line 241: Comment should be updated.

Jasper-Bekkers (changes requested):
We should also update `README.md` to reflect that OpSwitch works on 64-bit values now.

Issue #96, opened 2019-08-20
`Assemble::assemble(&self) -> Vec<u32>` always allocate heap memory


liigo:

Skepfyr:
Could you expand on this? I would have thought that assembly would almost always need to allocate somewhere, although I guess the return type could be made generic. What is the motivation here?

kvark:
My reading is that we shouldn't take responsibility of allocation here. Instead, we should either accept an existing `&mut Vec<>` or return an iterator, allowing the users to allocate.

paulkernfeld:
I see that there are a lot of uses of `Vec` throughout `rspirv` crate. So, even if `Assemble::assemble` didn't allocate, the crate would still need to make a lot of allocations internally. @liigo would this still be helpful to you?

liigo:
yes, thank you.

Jasper-Bekkers:
I think  #165 fixes this.

PR #165, opened 2020-09-16
Remove a lot of `vec!` macro invocations from Assemble 

Jasper-Bekkers:
This adds a `assemble_into` which just produces SPIR-V into a `Vec<u32>` instead of making lots and lots of new ones, this should reduce the assemble functions heap traffic quite a bit. This should address #96 and help me out at the same time.

### DIFF
--- a/rspirv/binary/assemble.rs
+++ b/rspirv/binary/assemble.rs
@@ -3,146 +3,152 @@ use std::convert::TryInto;
 
 /// Trait for assembling functionalities.
 pub trait Assemble {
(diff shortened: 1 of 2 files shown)


khyperia (approved):
Would be nice to take a `std::io::Write` instead of a `Vec`, but this is better than before, so :heavy_check_mark:

Jasper-Bekkers:
I thought of that - but `io::Write` (or `io::Cursor`) mostly operates on bytes so we'd just be doing a bunch of back and forth between u32 and u8's. Since SPIR-V is naturally expressed as u32's, this seemed like the easier / more ergonomic option.

PR #166, opened 2020-09-17
binary: parser: Make ConsumerError contain dyn Error + Send

MarijnS95:
In one of our projects we propagate errors out of a thread using `thiserror` to nest them recursively. The enum `State` ends up in that error type too, hence needs to be Send. Fortunately only `StringError` seems to be stored in this field currently.


Jasper-Bekkers (approved):
Looks good to me.

PR #167, opened 2020-09-24
impl trait is cooler than generics, use it

khyperia:
This _is_ a breaking change: see `rspirv/binary/disassemble.rs` `mod test` for an example of how it breaks.

I'm doing this primarily because the IDE experience in e.g. vscode rust-analyzer is significantly better: when autocompleting a method, the type of `impl trait` is directly in the position of the relevant parameter, and easily readable. With generics, you have to manually check the generic name, then look over to the generic parameter list, find the parameter, and check the type. This extra step has tripped me up enough times and caused me enough development pain that I got frustrated enough to write this PR.

### DIFF
--- a/autogen/build.rs
+++ b/autogen/build.rs
@@ -16,13 +16,13 @@ use std::{
 };
 use utils::write_autogen_comment;
 
-fn write<T: ToString>(path: &PathBuf, contents: T) {
+fn write(path: &PathBuf, contents: impl ToString) {
     let mut f = fs::File::create(path).expect(&format!("cannot open file: {:?}", path));
     write_autogen_comment(&mut f);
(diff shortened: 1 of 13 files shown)


Jasper-Bekkers (approved):

Jasper-Bekkers:
I think this is fine - always good to have productivity improvements. 

Would be good to get some extra eyes on since this is indeed a backcompat breaking change.

kvark (approved):
I didn't realize it makes a difference to IDE.

khyperia:
Screenshots to show the difference - these are popups when hovering over a `type_function` call. (There's still work to do in rust-analyzer to like... actually insert spaces here... but yeah)
image
image

PR #168, opened 2020-10-05
Perf: Don't allocate as much in Decoder::string

khyperia:
On a sample module, time taken _just_ in the Decoder::string method was 15% of the entire parse time, a significant majority of which is in the `extend` call reallocating over and over. This is obviously a hot spot, so it should be optimized a little more.

So, instead of consuming words and building up a buffer, do a scan over the bytes, extract the string, and advance the proper length. This does only one allocation, the `.to_string` at the end, as opposed to the log(n) allocations before.

This new method is only 5% of the entire parse time. Parse time for the module went from `14.866765ms` to `12.630514ms`, a 15% improvement. 1% of that 5% time is scanning for the null byte, another 1% for utf8 validation, and 3% for the `to_string` allocation/copy.

(Unfortunately there is no std function for null byte scan and utf8 validation at the same time, which would speed this up even more :cry:. Also, there's an optimized `memchr` function in std, but frustratingly, it's not exposed anywhere - it's *so close* right here, but they discard the position :( https://github.com/rust-lang/rust/blob/efbaa413061c2a6e52f06f00a60ee7830fcf3ea5/library/core/src/slice/cmp.rs#L273)

### DIFF
--- a/rspirv/binary/decoder.rs
+++ b/rspirv/binary/decoder.rs
@@ -3,6 +3,7 @@ use crate::spirv;
 use super::DecodeError as Error;
 use std::convert::TryInto;
 use std::result;
+use std::str;
 
 pub type Result<T> = result::Result<T, Error>;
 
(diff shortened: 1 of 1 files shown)


rspirv/binary/decoder.rs
```diff
@@ -160,20 +161,31 @@ impl<'a> Decoder<'a> {
...
+            None => &self.bytes[self.offset..],
+        };
+        // Find the null terminator.
+        let first_null_byte =
+            slice
+                .iter()
+                .position(|&c| c == 0)
+                .ok_or_else(|| match self.limit {
+                    Some(_) => Error::LimitReached(self.offset + slice.len()),
+                    None => Error::StreamExpected(self.offset),
```
Jasper-Bekkers, line 176: Shouldn't this also be self.offset + slice.len() since previously this would only happen scanning one of the last words in the slice?

Jasper-Bekkers (approved):

rspirv/binary/decoder.rs
khyperia: No, tests assert this behavior. I don't know why, but, they do.

PR #169, opened 2020-10-13
Add helper methods I'm constantly redefining

khyperia:
1) `Operand::unwrap_*`, mostly for `unwrap_id_ref`, but also `unwrap_literal_u32` and friends is useful. Just autogenerated all of them for completeness.
2) `Operand::unwrap_id_ref_any`, useful for rewriting any and all references to another ID
3) `Function::def_id` and `Block::label_id`, the usual chain of `func.def.as_ref().unwrap().result_id.unwrap()` is painfully long.
4) `Function::all_inst_iter{,_mut}`, surprisingly useful and relevant in various places.

### DIFF
--- a/autogen/dr.rs
+++ b/autogen/dr.rs
@@ -3,7 +3,7 @@ use crate::utils::*;
 
 use heck::SnakeCase;
 use proc_macro2::TokenStream;
-use quote::quote;
+use quote::{format_ident, quote};
 
 /// Returns true if the given operand kind can potentially have additional
 /// parameters.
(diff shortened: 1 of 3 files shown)


MarijnS95:
Oh this might have been very interesting for us in `rspirv-reflect!

khyperia:
> would that make sense here?

Hmm, maybe? The use case here is basically "I have checked `inst.class.opcode`, therefore I know the types of `inst.operands` for a fact, and if the types don't match up, that's a bug (panic).

In almost all cases where I imagine that I would conditionally check the operand type, it's where I want to handle the not-matching case immediately, and using `if let Operand::blah` seems to fit that well (and is more clear to me than chaining off an Option/Result return type).

khyperia:
Oh, actually, the `unwrap_id_ref_any` methods should return Option, because those are kind of a completely different use case, whoops

Jasper-Bekkers (approved):

MarijnS95:
@khyperia I don't recall if `rspirv` uses `Result` often, if not then this seems fine especially when parsing this specific operand format which should be rather set-in-stone I guess.

In our usecase we want to eliminate as many (all, if possible) panics, no matter how unlikely they might be (like encountering invalid spirv). That way the caller can decide whether to panic regardless or handle such a failure gracefully.

PR #171, opened 2020-11-12
Add FromStr impl for operands

khyperia:
This would be a pain to implement all the boilerplate for outside of rspirv, so throwing it in here for our use case in rust-gpu.

Note that this does _not_ implement FromStr for `bitflags!` operands, but implementing FromStr for those becomes a bit opinionated - do you create a list of values? how do you separate each individual value? so it's not really possible to impl it in an unopinionated way (and besides, there's not many bitflags, it's easy enough for users to implement their own FromStr).

### DIFF
--- a/autogen/header.rs
+++ b/autogen/header.rs
@@ -103,10 +103,12 @@ fn gen_value_enum_operand_kind(grammar: &structs::OperandKind) -> TokenStream {
     let mut from_str_impl = vec![];
     for e in &grammar.enumerants {
         if let Some(discriminator) = seen_discriminator.get(&e.value) {
+            let name_str = &e.symbol;
             let symbol = as_ident(&e.symbol);
             aliases.push(quote! {
                 pub const #symbol: Self = Self::#discriminator;
             });
+            from_str_impl.push(quote! { #name_str => Ok(Self::#discriminator), });
         } else {
             // Special case for Dim. Its enumerants can start with a digit.
             // So prefix with the kind name here.
--- a/autogen/header.rs
+++ b/autogen/header.rs
@@ -105,7 +105,7 @@ fn gen_value_enum_operand_kind(grammar: &structs::OperandKind) -> TokenStream {
         if let Some(discriminator) = seen_discriminator.get(&e.value) {
             let symbol = as_ident(&e.symbol);
             aliases.push(quote! {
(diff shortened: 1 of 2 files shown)


autogen/header.rs
```diff
@@ -109,18 +110,19 @@ fn gen_value_enum_operand_kind(grammar: &structs::OperandKind) -> TokenStream {
...
             } else {
                 e.symbol.to_string()
             };
-            let name = as_ident(&name);
+            let name = as_ident(&name_str);
             let number = e.value;
             seen_discriminator.insert(e.value, name.clone());
             enumerants.push(quote! { #name = #number });
             from_prim_list.push(quote! { #number => #kind::#name });
+            from_str_impl.push(quote! { #name_str => Ok(#kind::#name), });
```
MarijnS95, line 125: Could this be

```suggestion
            from_str_impl.push(quote! { #name_str => Ok(Self::#name), });
```

To save quite a bunch of characters on every line?

MarijnS95 (approved):
Awesome! Approval from me, if counts at all :sweat_smile: 

It's sad there's no `#[derive(FromStr)]` though, these are about ~800 extra lines for strings that appear identical to their `enum` variant yet could be generated at compile-time instead of checked in. The perks of auto-generated files :grin:

autogen/header.rs
```diff
@@ -100,27 +100,29 @@ fn gen_value_enum_operand_kind(grammar: &structs::OperandKind) -> TokenStream {
...
                 e.symbol.to_string()
             };
-            let name = as_ident(&name);
+            let name = as_ident(&name_str);
             let number = e.value;
             seen_discriminator.insert(e.value, name.clone());
             enumerants.push(quote! { #name = #number });
-            from_prim_list.push(quote! { #number => #kind::#name });
+            from_prim_list.push(quote! { #number => Self::#name });
+            from_str_impl.push(quote! { #name_str => Ok(Self::#name), });
```
Jasper-Bekkers, line 125: I think the aliases should probably also appear. Right now `ClosestHitKHR` isn't recognised but `ClosestHitNV` is for example, however the new way forward it through the KHR extension.

Jasper-Bekkers (changes requested):

Jasper-Bekkers (approved):

PR #173, opened 2020-11-20
autogen: Convert from build script to runnable binary

MarijnS95:
According to the autogen README rspirv-autogen is used to "automatically update files in anther crates". This use seems long-gone as none of the top-level projects depend on `rspirv-autogen`: regeneration does not happen at build-time anymore.

However, when loading this project in an IDE that scans and `check`s all workspaces the build script _is_ invoked resulting in slow startup times and accidental regeneration, which is particularly inconvenient when SPIRV-Headers so happens to be checked out on a different commit.

With this PR the recommended way to regenerate is:
```console
you@your-pc:~/rspirv $ cargo r -p rspirv-autogen
     Running `target/debug/rspirv-autogen`
```

In addition regeneration happens in the CI now as well, to capture any accidental mismatches between the generator, submodule checkout and submitted code.

(Edit: no, I don't like ```console either.)


Jasper-Bekkers:
I think this is a change for the better, would be nice to document the manual step in one of the README.md files so it's clear what to do for new (and old) contributors.

Jasper-Bekkers (approved):

PR #174, opened 2020-11-23
Regenerate with final KHR Ray Tracing spec

MarijnS95:
Depends on #173 

The ray tracing specification is final :tada: ! See https://www.khronos.org/blog/vulkan-ray-tracing-final-specification-release for more details.

We should hold off merging and releasing this until `SPIRV-Headers` drops the next tag so that we can version the `spirv_headers` crate properly.


MarijnS95:
Turned `TerminateRayKHR`/`IgnoreIntersectionKHR` into terminals (see https://github.com/microsoft/DirectXShaderCompiler/issues/3285, https://github.com/microsoft/DirectXShaderCompiler/pull/3295, https://github.com/KhronosGroup/SPIRV-Tools/pull/4050). Note that rspirv doesn't yet make the distinction between block terminals and execution terminators (these instructions + `Kill` + `TerminateInvocation`), is that something worth looking into?

Also removed the `Terminator` class. In particular because it's not possible to write `Some(Terminator) | (Some(Reserved) if ...)` in a match block, but this class has never existed in the history of `SPIRV-Headers`. Bit misleading to keep it in until it actually shows up, right?

Finally, the version: Looks like we have `1.5.4`, `1.5.4.raytracing` and `1.5.4.raytracing.fixed`... We can't represent that in semver and I have no idea if another proper version _number_ bump is going to happen to land the RT spec (here's hoping these are "prerelease" indicators...).

PR #175, opened 2020-11-27
sr: Add ray-tracing types and deduce OpPhi type in a consistent way

MarijnS95:
While randomly trying out `sr` to see what data would come out of it I ran into these issues, thought it was only fair to submit the fixes :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,9 +23,9 @@ checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
 name = "assert_matches"
-version = "1.4.0"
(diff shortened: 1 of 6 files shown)


khyperia (approved):
Not sure about the OpPhi change because I don't use SR (mostly wanting the content of #187 to be merged), but it looks sane to me!

MarijnS95:
@khyperia Thanks! I can move the OpPhi change to a separate PR since squash-merging would make a mess of that history anyway :)

khyperia:
eh, should be fine! pretty sure the change is okay, I just don't know the quirks of SR and can't 100% say it's correct, just 95% :P

MarijnS95:
Was already moving it over to a separate PR :)

PR #183, opened 2021-01-15
Add extension information to instructions

khyperia:
The spirv spec JSON has, for example, this information:

```json
{
  "opname" : "OpDemoteToHelperInvocationEXT",
  "class"  : "Reserved",
  "opcode" : 5380,
  "capabilities" : [ "DemoteToHelperInvocationEXT" ],
  "extensions" : [ "SPV_EXT_demote_to_helper_invocation" ],
  "version" : "None"
},
```
```json
{
  "enumerant" : "DemoteToHelperInvocationEXT",
  "value" : 5379,
  "capabilities" : [ "Shader" ],
  "extensions" : [ "SPV_EXT_demote_to_helper_invocation" ],
  "version" : "None"
},
```

This exposes the `extensions` json field in both `inst.class.extensions`, as well as, for example, `capability.required_extensions()` (both those two types of JSON extension information). This is useful for various things in rust-gpu.

Pretty much just blindly copy-pasted the existing generation code for `capability` except swapped it out for strings instead. Handily, we don't ever have to allocate a `String`, through the magic of `&'static str`.


Jasper-Bekkers:
Looks alright - just to sanity check `required_extensions` can return several extensions, for example

```
             | BuiltIn::IncomingRayFlagsNV => &["SPV_NV_ray_tracing", "SPV_KHR_ray_tracing"],
```
And so the result of this should be interpreted as "either SPV_NV_ray_tracing, or SPV_KHR_ray_tracing" (both wouldn't make sense).

khyperia:
yeah, it's really unclear to me in the spec whether there's an AND or an OR in some of these places - does `OpPtrDiff` require `Addresses` AND `VariablePointers`, or OR? so I just did the same ambiguous confusion that's already there and just copied in the JSON data directly, haha.

Jasper-Bekkers (approved):

Jasper-Bekkers:
I think it's OR in that case as well.

PR #185, opened 2021-02-11
Additional operands support (from BitEnum and ValueEnum)

Jasper-Bekkers:
Previously in PR #183  we added required_capabilities and required_extension to spirv_headers. This PR moves those into `dr::Operand` and adds `additional_operands`.

`additional_operands` is needed because some enum operands to instructions can specify additional operands that need parsing. Case in point would be an `OpDecorate` which only has one operand by default and the remaining operands are specified through that. So to properly parse `OpDecorate DescriptorSet 0` one needs to have called this `additional_operands` function to know what follows after `DescriptorSet`.


khyperia (approved):

kvark:
It's unfortunate that `required_xxx` now requires `rspirv` itself rather than the headers.

PR #186, opened 2021-03-02
Expose `iter` method to iterate over all instructions

Jasper-Bekkers:
This helped to speed up the compilation times of rust-gpu which uses rspirv in it's inline assembler.


autogen/src/dr.rs
```diff
@@ -291,12 +315,12 @@ pub fn gen_dr_operand_kinds(grammar: &[structs::OperandKind]) -> TokenStream {
                 }
             }
         });
-        let unwraps = kind_and_ty.into_iter().map(|(kind, ty)| {
+        let unwraps = kind_and_ty.iter().map(|(kind, ty)| {
```
MarijnS95, line 318: `kind_and_ty` doesn't appear to be used after this line (ie. might be a development artifact), a move with `into_iter` and without the extra `clone()` below should be faster?

autogen/src/dr.rs
```diff
@@ -308,6 +332,221 @@ pub fn gen_dr_operand_kinds(grammar: &[structs::OperandKind]) -> TokenStream {
...
+                            structs::Category::BitEnum => {
+                                use heck::ShoutySnakeCase;
+
+                                as_ident(&e.symbol.to_shouty_snake_case().replace("NA_N", "NAN"))
+                            }
+                            structs::Category::ValueEnum => {
+                                let name_str = if kind == "Dim" {
+                                    let mut name = "Dim".to_string();
+                                    name.push_str(&e.symbol);
+                                    name
```
MarijnS95, line 362: I wonder how much faster this is in than `format("Dim{}", &e.symbol)`. It's been a while since I last went through Rusts formatting internals, but that should expand to `.push_str` as well and `Display::fmt` on a string reference like `&e.symbol` should be a cheap reference.

autogen/src/dr.rs
```diff
@@ -308,6 +332,221 @@ pub fn gen_dr_operand_kinds(grammar: &[structs::OperandKind]) -> TokenStream {
...
+
+                let mut capability_clauses = BTreeMap::new();
+                let mut extension_clauses = BTreeMap::new();
+                let mut operand_clauses = BTreeMap::new();
+
+                let kind = as_ident(kind);
+                let mut seen_discriminator = BTreeMap::new();
+
+                for e in enumerators {
+                    if seen_discriminator.get(&e.value).is_none() {
```
MarijnS95, line 351: pedantic: you're never reading the name (RHS) of `seen_discriminator`, this could be a `BTreeSet` instead. Then also note that `.insert()` returns `true` if it managed to insert the item, `false` if it already exists. Could be a simple `if seen_descriminator.insert(e.value) { ... }`.

MarijnS95 (commented):
Looks like I'm too late. It's just generator comments anyway :shrug:

PR #187, opened 2021-03-31
Add support for generating OpTypeAccelerationStructureKHR

XAMPPRocky:
This PR adds support for `OpTypeAccelerationStructureKHR` by adding a small pass to manually map reserved functions into another class, so that the methods for those ops can be properly generated.

See also https://github.com/EmbarkStudios/rust-gpu/pull/563 which uses this change.

### DIFF
--- a/autogen/src/main.rs
+++ b/autogen/src/main.rs
@@ -44,7 +44,9 @@ fn map_reserved_instructions(grammar: &mut structs::Grammar) {
         .filter(|i| i.class == Some(structs::Class::Reserved))
     {
         match &*instruction.opname {
-            "OpTypeAccelerationStructureKHR" => instruction.class = Some(structs::Class::Type),
+            | "OpTypeAccelerationStructureKHR"
+            | "OpTypeRayQueryKHR"
+                => instruction.class = Some(structs::Class::Type),
             _ => {}
         }
     }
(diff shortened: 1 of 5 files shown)


MarijnS95:
This looks very close to what is done in #175, albeit that has more types. How about we review and merge that first?

XAMPPRocky:
> This looks very close to what is done in #175, albeit that catches every reserved OpType. How about we review and merge that first?

#175 as far as I can tell doesn't add a `Builder::type_acceleration_structure_khr` method, which is the important part of this PR that we need for generating those types in `rust-gpu`.

MarijnS95:
> #175 as far as I can tell doesn't add a `Builder::type_acceleration_structure_khr` method, which is the important part of this PR that we need for generating those types in `rust-gpu`.

I can pick that on top if you want, or rebase this PR to introduce that separately. At least we'll have more generic handling for reserved `OpType` instructions don't you think?

XAMPPRocky:
> I can pick that on top if you want, or rebase this PR to introduce that separately. At least we'll have more generic handling for reserved OpType instructions don't you think?

Anything that makes that method available works for me.

MarijnS95:
Allright, merged both solutions!

I like your approach more because it sets the type properly for every gen step, to add the missing method through `gen_dr_builder_types`. I hope that won't break things elsewhere when future changes come around, not sure what the deal is with these "reserved"-these-are-a-type-but-not-really classes anyway :thinking:

PR #193, opened 2021-05-05
Implement partial OpLine support

khyperia:
This implements partial OpLine support. "Partial" because this only supports OpLine in two places: in types_global_values, and inside basic blocks. In reality, OpLine is allowed in types_global_values, inside basic blocks, but also between basic blocks, in the OpFunctionParameter header, between functions, after all functions, and all sorts of other wacky locations. Supporting all those wacky locations would take a fairly significant refactor (for example, adding a `Vec<Instruction>` in basic blocks for "OpLine/OpNoLines preceding the OpLabel instruction" - because yes, there can be multiple, if we want to faithfully represent all valid spir-v modules)

Turns out, though, this partial support is kind of all that we need in rust-gpu, we don't need to represent wacky shenanigans, so I'll leave the full refactor work up to whoever needs that functionality in the future :)

Additionally, the debug section needs to be split up into its three component subsections to not intermix OpString and OpName instructions (which, OpString is relevant to OpLine because that's how the source file is specified). The spec states:

> 7. These debug instructions, which must be grouped **in the following order**:
>    a. All OpString, OpSourceExtension, OpSource, and OpSourceContinued, without forward references.
>    b. All OpName and all OpMemberName.
>    c. All OpModuleProcessed instructions.

I don't know why they didn't just make it three distinct "top-level" sections, because that's essentially what these subsections are, and it just causes a lot of confusion. Anyway, unfortunately, splitting up the subsections is a breaking change, so reviewers should be aware of that and consider alternatives (but I think a breaking change is the right way to go here).

### DIFF
--- a/rspirv/dr/build/mod.rs
+++ b/rspirv/dr/build/mod.rs
@@ -603,10 +603,10 @@ impl Builder {
         Ok(_id)
     }
 
-    /// Appends an OpLine instruction.
+    /// Appends an `OpLine` instruction.
     ///
-    /// If a block is currently selected, the OpLine is inserted into that block. If no block is
-    /// currently selected, the OpLine is inserted into types_global_values.
+    /// If a block is currently selected, the `OpLine` is inserted into that block. If no block is
+    /// currently selected, the `OpLine` is inserted into `types_global_values`.
     pub fn line(&mut self, file: spirv::Word, line: u32, column: u32) {
         let inst = dr::Instruction::new(
             spirv::Op::Line,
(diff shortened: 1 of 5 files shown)


rspirv/dr/build/mod.rs
```diff
@@ -602,6 +602,47 @@ impl Builder {
...
+            // types_global_values is the only valid section (other than functions) that
+            // OpLine/OpNoLine can be placed in, so put it there.
+            self.module.types_global_values.push(inst);
+        }
+    }
+
+    /// Appends an OpNoLine instruction.
+    ///
+    /// If a block is currently selected, the OpLine is inserted into that block. If no block is
+    /// currently selected, the OpLine is inserted into types_global_values.
```
MarijnS95, line 634:
```suggestion
    /// If a block is currently selected, the `OpNoLine` is inserted into that block. If no block is
    /// currently selected, the `OpNoLine` is inserted into `types_global_values`.
```

MarijnS95 (approved):

rspirv/dr/constructs.rs
```diff
@@ -34,8 +34,12 @@ pub struct Module {
...
     /// All execution mode declarations, using OpExecutionMode.
     pub execution_modes: Vec<Instruction>,
-    /// All non-location debug instructions.
-    pub debugs: Vec<Instruction>,
+    /// Debug subsection: All OpString, OpSourceExtension, OpSource, and OpSourceContinued.
+    pub debug_string_source: Vec<Instruction>,
+    /// Debug subsection: All OpName and all OpMemberName.
+    pub debug_names: Vec<Instruction>,
+    /// Debug subsection: All OpModuleProcessed instructions.
+    pub debug_module_processed: Vec<Instruction>,
```
eddyb, line 42: Tiny nit pick: looks like all the other names are pluralized, so I'd expect `debug_string_sources` and `debug_module_processeds`.

eddyb (approved):

PR #194, opened 2021-05-05
Make cargo clippy be clean

khyperia:
```
> cargo clippy &| wc -l
16723
```

was a little too much for me to cope with

A large chunk of this change is autogen changes, thankfully, not much "real" diff. Biggest thing is slapping `#[allow(clippy::upper_case_acronyms)]` all over.

### DIFF
--- a/autogen/src/utils.rs
+++ b/autogen/src/utils.rs
@@ -27,7 +27,7 @@ pub fn as_ident(ident: &str) -> Ident {
 /// given operand `kind` in the grammar.
 pub fn get_dr_operand_kind(kind: &str) -> Ident {
     as_ident(
-        if kind == "LiteralInteger" || kind == "LiteralContextDependentNumber" {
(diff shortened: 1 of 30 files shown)


autogen/src/utils.rs
```diff
@@ -26,14 +26,14 @@ pub fn as_ident(ident: &str) -> Ident {
...
-    as_ident(if kind == "LiteralInteger" {
-        "LiteralInt32"
-    } else if kind == "LiteralContextDependentNumber" {
-        // TODO: should use the correct type to decode
-        "LiteralInt32"
-    } else {
-        kind
-    })
+    as_ident(
+        if kind == "LiteralInteger" || kind == "LiteralContextDependentNumber" {
```
MarijnS95, line 30: `matches!` like you used elsewhere before?

MarijnS95 (approved):
Nice, that's some good cleanup :+1:!

PR #200, opened 2021-06-07
Add type_<name>_id builders

khyperia:
`type_struct_id` already existed, and these methods do the same thing as that, but for everything else too. This is useful when the user of rspirv does not want to invoke the (`O(n^2)`) `dedup_insert_type` method - they can pass in `Some(builder.id())` themselves and that case will be skipped.

### DIFF
--- a/autogen/src/dr.rs
+++ b/autogen/src/dr.rs
@@ -635,12 +635,30 @@ pub fn gen_dr_builder_types(grammar: &structs::Grammar) -> TokenStream {
         let name = as_ident(&inst.opname[2..].to_snake_case());
 
         let comment = format!("Appends an Op{} instruction and returns the result id, or return the existing id if the instruction was already present.", opcode);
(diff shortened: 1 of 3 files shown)


MarijnS95:
You could probably make `fn #name` call `fn #name_id` with `None` for `result_id` and save a ton of code and compile-time, or did I overlook a minor difference between the two?

khyperia:
Done!

> or did I overlook a minor difference between the two?

Nope, just laziness, calling through takes a bit of weird messing about with argument vs. parameter lists in the code generator, but I did that.

MarijnS95 (approved):

MarijnS95:
> Nope, just laziness, calling through takes a bit of weird messing about with argument vs. parameter lists in the code generator, but I did that.
Ah right, that's true - passing the parameters onwards is a bit awkward to fit in codegen, but it's nothing too bad :)

msiglreith (approved):

PR #203, opened 2021-07-05
Iterator -> IntoIterator

Matthias-Fauconneau:
Looks like taking Iterator instead of IntoIterator was a typo.
IntoIterator allows to directly use an array of operands as argument. 
And it is the trait used by extend (currently going through impl IntoIterator for Iterator).

### DIFF
--- a/rspirv/dr/build/mod.rs
+++ b/rspirv/dr/build/mod.rs
@@ -587,7 +587,7 @@ impl Builder {
         result_id: Option<spirv::Word>,
         extension_set: spirv::Word,
         instruction: spirv::Word,
-        operands: impl Iterator<Item = dr::Operand>,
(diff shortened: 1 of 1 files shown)


MarijnS95:
Note that PRs are squashed on merge, so it might make more sense (history-wise) to submit both commits individually instead of cramming it all into a single title (and commit).

.gitignore
```diff
@@ -3,3 +3,5 @@ target
 *.pyc
 *.bk
 .DS_Store
+autogen/external
+autogen_*.rs
```
Jasper-Bekkers, line 7: Intentional?

Jasper-Bekkers (approved):

Matthias-Fauconneau:
Moved

PR #204, opened 2021-07-08
Rename `spirv_headers` => `spirv` and reset to version `0.1.0+1.5.4`

### LINKED ISSUE #197: Switch spirv_headers to SemVer
See #117 and #196

Because spirv_headers doesn't follow SemVer, making changes to it, and keeping downstream dependencies from breaking we should switch over to SemVer and ditch the Vulkan / SPIR-V numbering scheme and instead just document which SPIR-V version corresponds to the current crate version. This should solve most of the problems in the future (but may introduce some confusion around which SPIR-V version is exposed instead).

I think it's probably better to have that confusion during development, rather then breaking downstream builds.

### LINKED ISSUE #202: Publish rspirv 0.8 ?
According to https://github.com/gfx-rs/rspirv/blob/master/rspirv/Cargo.toml , rspirv is at 0.8 , but the published crate is still stuck at 0.7. That doesn't seem right?

MarijnS95:
Fixes #197
Fixes #202


rspirv/lib.rs
```diff
@@ -1,5 +1,3 @@
-#![doc(html_root_url = "https://docs.rs/rspirv/0.7/")]
```
MarijnS95, line 1: I don't think this is still needed for docs.rs? This is the first time I'm seeing it...

README.md
```diff
@@ -28,14 +28,14 @@ Rust.
 Documentation
 -------------
 
-The current implementation supports SPIR-V 1.4 (Revision 1).
+The current implementation supports SPIR-V 1.5 (Revision 4).
```
MarijnS95, line 31: I really am not sure about this. https://www.khronos.org/registry/spir-v/ mentions `Revision 5` but we're on the `1.5.4.raytracing.fixed` tag. There is no `1.5.5` tag.

README.md
```diff
@@ -80,7 +80,7 @@ Building a SPIR-V module, assembling it, parsing it, and then disassembling it:
 
 ```rust
 extern crate rspirv;
-extern crate spirv_headers as spirv;
+extern crate spirv as spirv;
```
Jasper-Bekkers, line 83: No need to `as spirv` here.

rspirv/lib.rs
Jasper-Bekkers: It's not something I've seen other crates do either.

Jasper-Bekkers (approved):

README.md
MarijnS95: Ah yes. In true Rust 2018 fashion we don't need `extern crate` at all anymore.

Issue #209, opened 2021-08-01
LiftContext::convert struggles with linked multi-stage SPIRV


incertia:
The two test shaders are identical.
```glsl
#version 460
void main(){}
```

```
$ glslc test.vert -o vert.spv
$ glslc test.frag -o frag.spv
$ spirv-link vert.spv frag.spv -o test.spv
$ cargo run rspirv-test vert.spv frag.spv test.spv
vert.spv: success
frag.spv: success
thread 'main' panicked at 'Id 1 is already used', /home/incertia/.cargo/registry/src/github.com-1ecc6299db9ec823/rspirv-0.7.0/lift/storage.rs:42:35
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We have a simple test driver that attempts to convert modules.
```rust
use rspirv::dr::*;
use rspirv::lift::*;
use std::env;
use std::fs::read;

fn main() {
  for file in env::args().skip(1) {
    match read(&file) {
      Ok(spv) => match load_bytes(spv) {
        Ok(module) => match LiftContext::convert(&module) {
          Ok(module) => {
            println!("{}: success", file);
          }
          Err(e) => {
            println!("{}: {:?}", file, e);
          }
        },
        Err(e) => {
          println!("{}: {}", file, e);
        }
      },
      Err(e) => {
        println!("{}: {}", file, e);
      }
    }
  }
}
```
`spirv-dis` does not report anything out of the ordinary.
```
$ spirv-dis test.spv
; SPIR-V
; Version: 1.0
; Generator: Khronos; 17
; Bound: 8
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main"
               OpEntryPoint Fragment %main_0 "main"
               OpExecutionMode %main_0 OriginUpperLeft
               OpSource GLSL 460
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpSource GLSL 460
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %main_0 "main"
       %void = OpTypeVoid
          %5 = OpTypeFunction %void
       %main = OpFunction %void None %5
          %6 = OpLabel
               OpReturn
[...]

PR #210, opened 2021-08-01
apply the correct result type when lifting functions

incertia:
should address #209 but i'm not sure if what i did is entirely good

either way it's a 2 line change so it should be quick and easy


MarijnS95:
> as a side note it looks like 0.8.0 was not pushed to crates.io

See #204 and all the relevant linked issues, `SPIRV-Headers` caused a semver issue leading to 0.8.0 to be yanked. We're in the process of rectifying that and can hopefully merge most of the outstanding PRs before finally publishing the long-awaited crate updates ASAP :grimacing:

MarijnS95 (approved):
I'm not entirely familiar with this code either, but the end result looks sound. At least we're not appending a new "void" type for every function anymore which is a good thing.

msiglreith (approved):

PR #212, opened 2021-08-16
Fix release regex

Jasper-Bekkers:


rspirv/Cargo.toml
```diff
@@ -22,10 +22,7 @@ clippy = { version = "0.0", optional = true }
 derive_more = "0.99"
 fxhash = "0.2"
 num-traits = "0.2"
-
-[dependencies.spirv]
-version = "=0.1"
-path = "../spirv"
+spirv = { version = "=0.1.0+1.5.4", path = "../spirv" }
```
MarijnS95, line 25: A dependency shouldn't use the `+metadata` bit though, I hope `cargo-release` understands and strips that off?

spirv/release.toml
```diff
@@ -7,5 +7,7 @@ sign-tag = true
 
 pre-release-replacements = [
   {file="README.md", search="spirv = .*", replace="{{crate_name}} = \"{{version}}\""},
-  {file="../rspirv/Cargo.toml", search="spirv = .*", replace="{{crate_name}} = \"={{version}}\""},
-]
\ No newline at end of file
+  {file="../rspirv/Cargo.toml", search="spirv = \\{ version = \"=.*\", path = \"../spirv\" \\}", replace="{{crate_name}} = { version = \"={{version}}\", path = \"../spirv\" }" },
```
MarijnS95, line 10: Not sure if we should keep the `=` sign; it means that `rspirv` is fixed to spirv `0.1.0` exactly. If we ever release non-breaking spirv changes in `0.1.1` (which is what we want to do now with the semver restart) none of the `rspirv` crates will be able to use it unless they get a bump too.


(I initially added the `=` but don't recall why, since as explained above it obviously doesn't make much sense).

rspirv/Cargo.toml
Jasper-Bekkers: It ouput's like this which seems to make sense: `Fixing rspirv's dependency on spirv to '= 0.1.1' (from '= 0.1.0')"`

spirv/release.toml
Jasper-Bekkers: I'll remove it.

rspirv/Cargo.toml
MarijnS95: @Jasper-Bekkers Does that mean it's already bumping the version to 0.1.1 before releasing? We should teach `cargo-release` that `0.1.0` is our first release that isn't actually out yet. Maybe we should set it to `0.0.0` and pass `0.1.0` or `1.0.0` as argument to `cargo-release` instead?

Jasper-Bekkers: Good point! :-)

MarijnS95 (approved):

PR #216, opened 2021-09-23
Unify extended instruction parsing and add `NonSemantic.DebugPrintF`

Matthias-Fauconneau:


autogen/src/main.rs
```diff
@@ -61,65 +62,82 @@ fn main() {
...
+            serde_json::from_str(
+                &std::str::from_utf8(
+                    &fs::read(autogen_src_dir.join(
+                        "external/SPIRV-Headers/include/spirv/unified1/spirv.core.grammar.json",
+                    ))
+                    .unwrap(),
+                )
+                .unwrap(),
+            )
+            .unwrap();
```
MarijnS95, line 76:
```suggestion
        let mut original = serde_json::from_str(
            &fs::read_to_string(
                autogen_src_dir
                    .join("external/SPIRV-Headers/include/spirv/unified1/spirv.core.grammar.json"),
            )
            .unwrap(),
        )
        .unwrap();
```

Maybe keep this simple with `read_to_string`? Same below.

autogen/src/main.rs
```diff
@@ -61,65 +62,82 @@ fn main() {
...
-        );
-        let mut file = fs::File::open(path).unwrap();
-        contents.clear();
-        file.read_to_string(&mut contents).unwrap();
-    }
-    let cl_grammar: structs::ExtInstSetGrammar = serde_json::from_str(&contents).unwrap();
+    let extended_instruction_sets = [
+			("GLSL.std.450", "GLOp", "https://www.khronos.org/registry/spir-v/specs/unified1/GLSL.std.450.html"),
+			("OpenCL.std.100", "CLOp", "https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.ExtendedInstructionSet.100.html"),
+			("NonSemantic.DebugPrintF", "DebugPrintFOp", "https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/debug_printf.md"),
```
MarijnS95, line 84: Please use spaces to indent this. The autoformatter refuses to do so because the lines are too long.

```suggestion
        ("GLSL.std.450", "GLOp", "https://www.khronos.org/registry/spir-v/specs/unified1/GLSL.std.450.html"),
        ("OpenCL.std.100", "CLOp", "https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.ExtendedInstructionSet.100.html"),
        ("NonSemantic.DebugPrintF", "DebugPrintFOp", "https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/docs/debug_printf.md"),
```

autogen/src/main.rs
Matthias-Fauconneau: Yes! I didn't know about it ><.
There really should be a note in std:fs::read about read_to_string existence. It's not the first time this annoys me ^^

PR #219, opened 2021-11-16
Remove the num-traits dependency

bjorn3:
It is only used for the FromPrimitive trait which could trivially be replaced with from_u32 functions. This gives a compile time win of ~1.8s for the spirv crate. (~3.8s -> ~2.0s)


MarijnS95 (approved):

MarijnS95:
Glad to see this go in addition to `derive-more`!

Jasper-Bekkers:
Glad to see these terrible crates go :-)

bjorn3:
That's quick @MarijnS95!

PR #221, opened 2021-11-16
Switch to rustc-hash

bjorn3:
The last fxhash version is 4 years old while the last rustc-hash version is 2 years old. Rustc-hash is an official rust project unlike fxhash. Rustc-hash doesn't have any dependencies, while fxhash has a dependency on byteorder. Finally rustc-hash takes ~0.7s to compile while fxhash including it's byteorder dependency takes ~1.0s to compile.

### DIFF
--- a/rspirv/Cargo.toml
+++ b/rspirv/Cargo.toml
@@ -19,7 +19,7 @@ travis-ci = { repository = "gfx-rs/rspirv" }
 
 [dependencies]
 clippy = { version = "0.0", optional = true }
-fxhash = "0.2"
+rustc-hash = "1.1.0"
 spirv = { version = "0.2.0", path = "../spirv" }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


MarijnS95 (approved):
Nice @bjorn3, you're really having at it!

Jasper-Bekkers (approved):

Jasper-Bekkers:
@bjorn3 if you need a new release (and are done with these changes) let me know in an issue and I'll spin one.

a1phyr:
> I'm trying to reduce the compile time of wgpu.

You could also use `ahash`, which is already in the dependency tree through `hashbrown`.

bjorn3:
Naga doesn't enable the ahash feature of hashbrown though wgpu does through gpu-descriptors. By the way I see that gpu-descriptor uses an older hashbrown version than indexmap which is used by naga.

PR #222, opened 2021-11-17
Use transmute to convert numbers to enum variants

bjorn3:
This reduces the size of the generated matches. This saves about 1200 lines of generated code. In addition it reduces the amount of generated llvm ir from 20807 lines to 18881 lines. The object file size is reduced by 70kb. I haven't checked performance.

I also tried improving the size of the derived debug impls using a manual impl, but that resulted in more llvm ir for some reason.

I don't know of any other possible improvements to the `spirv` crate and I don't plan to try to improve compile times for `rspirv` as it isn't used in the default configuration of wgpu and bevy, so a new release would be nice @Jasper-Bekkers. #219 is a semver incompatible change, so it must be released as 0.3.0.


autogen/src/header.rs
```diff
@@ -43,9 +44,55 @@ fn bit_enum_attribute() -> TokenStream {
...
+    let enumerants = variants
+        .iter()
+        .map(|(number, name)| quote! { #name = #number });
+
+    let mut number_runs = vec![];
+    number_runs.push(variants[0].0..=variants[0].0);
+    for &(number, _) in variants.iter().skip(1) {
+        let last_run = number_runs.last_mut().unwrap();
+        if number == *last_run.end() + 1 {
+            *last_run = *last_run.start()..=*last_run.end() + 1;
```
MarijnS95, line 63: Just:
```suggestion
            *last_run = *last_run.start()..=number;
```
?

I wonder if there's a more concise way of writing this though.

autogen/src/header.rs
```diff
@@ -152,8 +198,7 @@ fn gen_value_enum_operand_kind(grammar: &structs::OperandKind) -> TokenStream {
             let name = as_ident(&name_str);
             let number = e.value;
             seen_discriminator.insert(e.value, name.clone());
-            enumerants.push(quote! { #name = #number });
-            from_prim_list.push(quote! { #number => Self::#name });
+            from_prim_list.push((number, name.clone()));
```
MarijnS95, line 201: We could rename this value to better match its intent?

autogen/src/header.rs
bjorn3: Unfortunately there isn't. `RangeInclusive` doesn't have any setters.

MarijnS95 (approved):
Overall this is a pretty clever and welcome change, I like it :)

autogen/src/header.rs
MarijnS95: Yeah, and keeping a min and max in a `u32` separately, updating the max bounds, and adding it whenever the next number is not contiguous _or we run out of numbers_ (not to mention resetting min and max) costs many more lines, too.

MarijnS95: Even though I'm a fan of `RangeInclusive` none of this code takes advantage of it (it's only used as `RangeInclusive` in the final generated code) so you might as well use a tuple for the min and max which you _can_ update in place?

autogen/src/header.rs
bjorn3: Done

autogen/src/header.rs
```diff
@@ -43,9 +44,55 @@ fn bit_enum_attribute() -> TokenStream {
...
+    for &(number, _) in variants.iter().skip(1) {
+        let last_run = number_runs.last_mut().unwrap();
+        if number == *last_run.end() + 1 {
+            *last_run = *last_run.start()..=number;
+        } else if number > *last_run.end() + 1 {
+            number_runs.push(number..=number);
+        } else {
+            unreachable!("Variants not sorted by discriminant");
+        }
+    }
```
MarijnS95, line 69:
```suggestion
    let mut number_runs = vec![(variants[0].0, variants[0].0)];
    for &(number, _) in variants.iter().skip(1) {
        let last_run = number_runs.last_mut().unwrap();
        if number > last_run.1 + 1 {
            number_runs.push((number, number));
        } else if number == last_run.1 + 1 {
            last_run.1 = number;
        } else {
            unreachable!("Variants not sorted by discriminant")
        }
    }
```

Using a tuple simplifies this all, and we can just use `.map(|(start, end)| {` below :)

autogen/src/header.rs
bjorn3: Like this?

autogen/src/header.rs
bjorn3: Done

Jasper-Bekkers:
Nice stuff, I thought of doing something similar before but never did it because of the odd numeric "ranges" that vulkan has. Good to see it tackled.

PR #223, opened 2021-11-17
ci: Run `clippy -Dwarnings` and use minimal toolchain profile

MarijnS95:


MarijnS95:
@msiglreith Can we get this in?

msiglreith (approved):
sure (:

autogen/Cargo.toml
```diff
@@ -6,6 +6,7 @@ authors = [
 	"Lei Zhang <antiagainst@gmail.com>",
 ]
 edition = "2018"
+rust-version = "1.58"
```
MarijnS95, line 9: 3 years later, I don't understand why I added this: `autogen` is the only crate that _is not_ published and shouldn't care about the version at all, since Rust developers contributing to this repository and specifically using the generator to regenerate its contents can easily use the latest `stable` at all times.

PR #224, opened 2021-11-17
Fix all clippy lints for Rust 1.56.1

MarijnS95:
These slipped in over time.  Fix them once now, while enabling clippy in the CI separately to make sure they don't end up in the codebase over time either.


MarijnS95:
https://github.com/gfx-rs/rspirv/runs/4242859869?check_suite_focus=true

Jasper-Bekkers (approved):

PR #226, opened 2021-11-22
Regenerate with SPIRV-Headers sdk-1.2.198

MarijnS95:


MarijnS95:
I really don't remember if we should update `is_type` in `rspirv/grammar/reflect.rs` or why this is not autogenerated from `class: Type-Declaration`?

msiglreith:
I noticed that `OpSamplerImageAddressingModeNV` is treated as 'normal' instruction as it's tagged `reserved` (like all extension ops I guess)

MarijnS95:
@msiglreith I think with "module header" you are referring to `types_global_values`?

It seems this operand best fits the constants, it's not really a type but simply representing a constant `Bit Width`. Adding yet another exception for that was easy enough.

Any other such ops I should be aware of? Seems there's a lot of new spec to go through :(

MarijnS95:
Actually, a type is better since this operand doesn't have a `result_type` which is common/required across constants.

msiglreith:
> Any other such ops I should be aware of? Seems there's a lot of new spec to go through :(

Looking trough the latest extensions since the raytracing stuff, I didn't see something requiring special attention so far.

> Actually, a type is better since this operand doesn't have a result_type which is common/required across constants.

IMO this instruction requires a bit of special treatment similar to `OpMemoryModel` for example. Not sure where this instruction would land when parsing the reencoding it using rspirv if it's part of the type list. Removing it from the normal instruction list seems to good enough to me - I can make also a PR once this lands as the extensions is quite niche anyway

MarijnS95:
@msiglreith Looks like `OpMemoryModel` is implemented by hand in `rspirv/dr/build/mod.rs` because all `Mode-Setting` (`Class:Mode-Setting`) instructions are ignored.

You're right that this requires more changes - seems I only updated the data-representation builder part but not the loader/parser in `fn consume_instruction()`. I can ignore it from there (ie. write a `todo!()` to panic on this instead of doing the wrong thing) then you take it over in a separate PR? Will have to see what to do with the builder though. Alternatively I can skip `Consider OpSamplerImageAddressingModeNV to be a constant` and let you deal with and test the entire thing?

MarijnS95:
@msiglreith Completely ignoring the type for now. Shall we get #223 in and then get this PR on top?

msiglreith:
Does this pass the clippy checks introduced in #223?

msiglreith (approved):

MarijnS95:
@msiglreith Yeah it does, I just wanted to make sure that both PRs were merged separately :)

PR #228, opened 2021-12-27
Remove leading `///` in some auto-generated docs

LeSeulArtichaut:
Leading `///` are not trimmed in `#[doc = "..."]` and will appear in the generated documentation.


MarijnS95 (approved):
We should probably get this in ^^

PR #207, opened 2021-07-27
OpSelectionMerge is not a terminator

Matthias-Fauconneau:

PR #229, opened 2022-03-08
Fix OpSelectionMerge and OpLoopMerge incorrectly being block terminators

BinaryWarlock:
This is a resurrection of #207, which hasn't been active since July of last year.
It builds upon that and also includes OpLoopMerge in it.

In current rspirv, OpSelectionMerge and OpLoopMerge are block terminators, which is incorrect. In fact, you literally cannot use either of them right now, because they must immediately precede branch instructions, which is not possible since you cannot create a branch outside of a block in rspirv, and the merge instructions already terminated the block. So they cannot actually be constructed.

This fixes that. I've tested it in my own code.

I'd like to humbly request a version bump after this is merged, so that I don't have to maintain a local copy, plus fixing the API is probably a good reason for a release.

### DIFF
--- a/autogen/src/dr.rs
+++ b/autogen/src/dr.rs
@@ -716,7 +716,7 @@ fn is_terminator_instruction(inst: &structs::Instruction) -> bool {
         ),
         Some(structs::Class::Branch) => !matches!(
             inst.opname.as_str(),
-            "OpPhi" | "OpLabel" | "OpSelectionMerge"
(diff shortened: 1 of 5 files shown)


MarijnS95 (approved):

BinaryWarlock:
@msiglreith Fixed the formatting and autogen test failures. The warnings / lines from the clippy lints aren't even touched by this PR, so they're not relevant to it.

Thanks.

PR #231, opened 2022-04-03
Fix disassembly of constant signed int literals.

### LINKED ISSUE #227: Disassembling constant literals is broken
Looking at the disassembler logic, it prints out operands without any context. So for constant literals, it just prints them as a sequence of `u32`. This doesn't work for negative integers, and it likely breaks on 64-bit types.

See [shader playground](https://shader-playground.timjones.io/c6b6f3ae081a8a79b75e917520ca6649):
> %int_n2 = OpConstant %int -2

Downloading the binary and running it through `dis` in this repo gives me:
> %5 = OpConstant  %4  4294967294

Trying to assemble it back into SPIR-V using `spirv-as` fails with:
> error: 29: 23: Integer 4294967294 does not fit in a 32-bit signed integer

supervacuus:
Fixes #227.

This approach is localised to the disassembler. No changes to the parser.

- removed duplication for operand rendering in the disassembly of instructions
- introduced type-tracker for constant literals to resolve for signs of integer operands
- implemented correct rendering of signed integer literals
- added very basic test that covers integer literals

### DIFF
--- a/rspirv/binary/disassemble.rs
+++ b/rspirv/binary/disassemble.rs
@@ -55,9 +55,9 @@ fn disas_join(insts: &[impl Disassemble], delimiter: &str) -> String {
         .join(delimiter)
 }
 
-fn disas_instruction<F>(
-    inst: &dr::Instruction, space: &str, disas_operands: F) -> String
-    where F: Fn(&Vec<Operand>) -> String
+fn disas_instruction<F>(inst: &dr::Instruction, space: &str, disas_operands: F) -> String
+where
+    F: Fn(&Vec<Operand>) -> String,
 {
     format!(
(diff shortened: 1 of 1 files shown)


supervacuus:
The remaining failing check is unrelated to my changes. I can fix the clippy issue in the decoder, but the generated duplicate match for `OpReportIntersection[KHR|NV]` should probably be solved in another PR.

teoxoy:
Just wanted to mention that this PR fixes the issue I had in naga https://github.com/gfx-rs/naga/commit/2cda40396ac53fd38697053ae150d1733e8730d7.

msiglreith (approved):
thanks!

PR #234, opened 2022-05-28
Add `Sync` trait bound to make compatible with anyhow

BeastLe9enD:
Hey, I had a problem with `rspirv-reflect`, where I couldn't use `anyhow::Result` in the following example:

```Rust
use std::fs::File;
use std::io::Read;
use anyhow::Result;
use rspirv_reflect::Reflection;

fn main() -> Result<()> {
    let mut file = File::open("test.vert.spv")?;

    let mut buffer = Vec::new();
    file.read_to_end(&mut buffer)?;

    let reflection = Reflection::new_from_spirv(&buffer)?;
    let descriptor_sets = reflection.get_descriptor_sets()?;

    println!("{:#?}", descriptor_sets);

    Ok(())
}
```

The reason this does not work is because `rspirv::binary::ParseState` does not implement `Sync`. So I changed from `Box<dyn error::Error + Send` to `Box<dyn error::Error + Send + Sync`, and and it keeps compiling so I think that's ok.

### DIFF
--- a/rspirv/Cargo.toml
+++ b/rspirv/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rspirv"
-version = "0.11.0+sdk-1.2.198"
+version = "0.11.1+sdk-1.2.198"
 authors = ["Lei Zhang <antiagainst@gmail.com>"]
 edition = "2018"
 
--- a/rspirv/Cargo.toml
+++ b/rspirv/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rspirv"
-version = "0.11.1+sdk-1.2.198"
+version = "0.11.0+sdk-1.2.198"
 authors = ["Lei Zhang <antiagainst@gmail.com>"]
 edition = "2018"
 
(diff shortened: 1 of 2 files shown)


MarijnS95:
@BeastLe9enD This supersedes https://github.com/Traverse-Research/rspirv-reflect/pull/24, right? As you can see, sometimes things are easier/cleaner to resolve, thanks :slightly_smiling_face:

BeastLe9enD:
@MarijnS95 Yes, the first version I suggested was really unholy, I definitely get that 😄

rspirv/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rspirv"
-version = "0.11.0+sdk-1.2.198"
+version = "0.11.1+sdk-1.2.198"
```
MarijnS95, line 3: Let's leave release bumps to the maintainers, please. They will also update this in the `README` when the time comes for it.

MarijnS95 (approved):
Nice solution, `Action::Error` - hence `State::ConsumerError` - are only ever constructed with a custom `struct ErrorString(&'static str);` hence this seems like a simple/clean/correct solution to me. Perhaps we can even simplify this all into just storing a `&'static str` and get rid of unnecessary complication, but that's for a separate PR/discussion.

rspirv/Cargo.toml
BeastLe9enD: Ok great, rolled back the version

msiglreith (approved):

PR #235, opened 2022-07-07
Fix some clippy errors

p3t3rix:
Wanted to fix some clippy errors to make CI/CD green again.
The clippy allow was added because extensions turning official still remain in the grammar.json and generate arms with the same opcode. Not sure if this should be fixed in another way but i think making CI green should be enough for the near future.
I just touched clippy errors that were thrown by ci/cd when it last ran (for example here https://github.com/gfx-rs/rspirv/runs/6061904152?check_suite_focus=true).

### DIFF
--- a/autogen/src/sr.rs
+++ b/autogen/src/sr.rs
@@ -515,7 +515,6 @@ pub fn gen_sr_code_from_instruction_grammar(
     };
 
     let lift_context = quote! {
-        #[allow(clippy::match_overlapping_arm)]
         impl LiftContext {
             pub fn lift_branch(
                 &mut self, raw: &dr::Instruction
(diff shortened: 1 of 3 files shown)


MarijnS95:
> The clippy allow was added because extensions turning official still remain in the grammar.json and generate arms with the same opcode. Not sure if this should be fixed in another way but i think making CI green should be enough for the near future.

I don't think we should hide these, but find a solution instead. It's not only the match arm but the duplicate enum variants may also be confusing/wrong. Rather have red CI long enough so that someone picks them up and fixes this, may actually be pretty trivial.

MarijnS95:
@p3t3rix There are still two other issues that this PR fixes, you can leave it open for those and remove the `#[allow(clippy::match_overlapping_arm)]`, then make an issue to address that separately.

p3t3rix:
Reverted the clippy allow and created an issue for the duplicated op-codes: https://github.com/gfx-rs/rspirv/issues/236

MarijnS95 (approved):
Thanks!

msiglreith (approved):

Issue #189, opened 2021-04-02
Floating-point literals should store bits in integers, not use host f32/f64.


eddyb:
Compilers should never rely on host floating-point support, as that can lead to host-specific behavioral subtleties (instead of behavior being deterministically tied to the target architecture/platform), and in many cases there's not enough to gain to even look into what guarantees certain host architectures/platforms may offer.

For `rustc` we ported LLVM's C++ `APFloat` library to Rust (for e.g. evaluating float operations in constant expressions), but `rspirv` itself only needs to store literals, not operate on the actual values (and consumers could choose how to interpret the bits of a floating-point value, any way they want).

I'm not sure how what the timeline for 0.8 is, but this would be a **breaking change**.

Also, it would help with #16.

cc @khyperia

Jasper-Bekkers:
@eddyb do you have any examples of where this is done wrong? Is it just the `constant_f32` functions and `LiteralFloat32` etc?

eddyb:
Right, and loading code like this: https://github.com/gfx-rs/rspirv/blob/719cf08e4af0436242707479e3509add5ec3d514/rspirv/binary/decoder.rs#L205-L219

Basically the types `f32` and `f64` should more or less not show up in `rspirv` (except maybe for `fmt::Debug`?).

PR #237, opened 2022-10-03
Store floats as bits

### LINKED ISSUE #16: Instruction: Derive Eq and Hash
hey, for a personal project I need to store Instructions in a HashMap.

Can these derives be added to Instruction?

wackbyte:
Closes #16, should help with #189

### DIFF
--- a/rspirv/binary/disassemble.rs
+++ b/rspirv/binary/disassemble.rs
@@ -1,8 +1,8 @@
 use crate::binary::tracker::Type;
-use crate::binary::tracker::Type::Integer;
+use crate::binary::tracker::Type::{Float, Integer};
 use crate::dr;
 use crate::dr::Operand;
-use crate::dr::Operand::{LiteralInt32, LiteralInt64};
+use crate::dr::Operand::{LiteralBit32, LiteralBit64};
 use crate::spirv;
 
 use super::tracker;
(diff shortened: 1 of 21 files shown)


rspirv/dr/build/mod.rs
```diff
@@ -749,36 +749,9 @@ impl Builder {
...
-            vec![dr::Operand::LiteralFloat64(value)],
-        );
-        self.module.types_global_values.push(inst);
-        id
-    }
-
-    /// Appends an OpConstant instruction with the given 32-bit integer `value`.
-    /// or the module if no block is under construction.
-    pub fn constant_u32(&mut self, result_type: spirv::Word, value: u32) -> spirv::Word {
+    pub fn constant_int32(&mut self, result_type: spirv::Word, value: u32) -> spirv::Word {
```
MarijnS95, line 754: Nitpicking: could we also name these `constant_32bit` and `constant_64bit`? Same for the other types, so that it's a bit more clear that it's a generic "array of bytes" rather than a (float interpreted as) integer.

That also matches the updated doc-comment on this function (and we could have a similar doc-comment on the `dr::Operand` enum variant).

MarijnS95 (approved):
Thanks! Can we possibly solve all the clippy lints for the base repo in a separate PR to make this one a bit easier to read?

rspirv/dr/build/mod.rs
wackbyte: Decided to go with `bit32` and `bit64` instead to avoid any names starting with numbers.

wackbyte:
I removed the clippy commit :+1:

wackbyte:
Sorry about that, just had some trouble with git while trying to rebase.

rspirv/dr/build/mod.rs
MarijnS95: Thanks! I didn't expect it to be such an intrusive change, but worth it IMO. Will now have to wait for a maintainer to say the same :)

rspirv/binary/disassemble.rs
```diff
@@ -197,10 +197,10 @@ fn disas_constant(inst: &dr::Instruction, type_tracker: &tracker::TypeTracker) -
     debug_assert_eq!(inst.operands.len(), 1);
     let literal_type = type_tracker.resolve(inst.result_type.unwrap());
     match inst.operands[0] {
-        LiteralInt32(value) => disas_instruction(inst, " ", |_| {
```
msiglreith, line 200: the operand won't be disassembled correctly in case of a floating point literal type due to the underlying `disas_literal_int` code which does some type checking. Should be extended to support `Float` types as well otherwise we print an empty string.

msiglreith (changes requested):

rspirv/binary/disassemble.rs
wackbyte: ah, my mistake. nice catch!

wackbyte: hmm, i looked into it and it looks like this might require something similar to [this module in `spirv-tools`](https://github.com/KhronosGroup/SPIRV-Tools/blob/4dbc66380dd63aabbd33c38198008449d0a5807a/source/util/hex_float.h) in order to properly deal with NaNs, infinity, 16-bit floats, etc. i think this is outside the scope of this PR, though, so I left a TODO comment and decided to just rely on the `Display` impls of `f32` and `f64` to handle most usages, as i believe this is what the behavior was before anyway.

wackbyte:
Okay, finally rebased properly.

msiglreith (approved):
lgtm, thanks for adding a testcase for the float disassembling!
clippy nags about the chosen `3.14` constants tho :)

wackbyte:
Alright, changed it to something else

PR #242, opened 2023-04-05
sr: Deduplicate `match` ranges like `spirv` header

MarijnS95:
Solve some clippy lints by getting rid of these aliases of overlapping `Op`s.  Unfortunately, just like in `spirv`, this biases towards the `NV` rather than `KHR` stabilized raytracing ops.

Note that in `spirv` the `Op`s are aliases in code rather than omitted altogether: don't think we can do that for the range match here though unless deducing whether the shader was compiled with the NV or KHR variant, even though the KHR variant is already stabilized for quite some time.

### DIFF
--- a/autogen/src/main.rs
+++ b/autogen/src/main.rs
@@ -43,13 +43,66 @@ fn map_reserved_instructions(grammar: &mut structs::Grammar) {
         .iter_mut()
         .filter(|i| i.class == Some(structs::Class::Reserved))
     {
-        if instruction.opname.starts_with("OpType")
-            // Ignore AccelerationStructureNV which has the same opcode as AccelerationStructureKHR
-            && instruction.opname != "OpTypeAccelerationStructureNV"
-        {
+        if instruction.opname.starts_with("OpType") {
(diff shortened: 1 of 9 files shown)


msiglreith (approved):

MarijnS95:
@msiglreith are you okay with this change, or should we see if we can prefer the `KHR` types here (and maybe also in the `spirv` crate)?

msiglreith:
@MarijnS95 `KHR` > `EXT` > Vendor specific would be great but the current approach seems also totally fine to me

MarijnS95:
@msiglreith definitely, it might just be a hassle to implement as all parts of the generator redo these kinds of evaluations on the parsed SPIR-V grammar :(

MarijnS95:
@msiglreith done now!

msiglreith (approved):
great thanks!
(huh so many vendors already)

MarijnS95:
> (huh so many vendors already)

Yup rather than only listing the few vendors with ops currently I just copied the whole thing from https://github.com/ash-rs/ash/pull/676.

PR #243, opened 2023-05-27
Bump `bitflags` to v2

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented for bitflag types as well as the output by Serde is now different.


autogen/src/header.rs
```diff
@@ -34,6 +34,7 @@ fn value_enum_attribute() -> TokenStream {
 
 fn bit_enum_attribute() -> TokenStream {
     quote! {
+        #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
MarijnS95, line 37: even though these were the original `derive` attributes with `bitflags 1`, is `(Partial)Ord` really meaningful/useful for any of these?

daxpedda: Probably not. They could be added on demand at a later point without a breaking change so I think there would be nothing wrong with removing them now.

Let me know!

MarijnS95: Yeah, let's remove them. I double-checked all definitions and none of them seem to have any sort of meaning in ordering.

daxpedda: Done.

msiglreith (approved):

daxpedda:
Uh, I guess removing `Ord` had an issue after all.
`Decoration::FPFastMathMode` requires it. Should I remove it from `Decoration` or add it back to the macro again?

msiglreith:
As `FPFastMathMode` is a bitmask with no defined partial/total order relation I would suggest removing `PartialOrd`/`Ord` from `Decoration` as well

daxpedda:
Done.

MarijnS95 (approved):

msiglreith (approved):
Thanks, beside the test failure the other CI issues seem unrelated, I'm gonna merge it and fix the CI parts in a separate PR

PR #247, opened 2023-11-18
Address CI failures and pretty-print bit enum operands

msiglreith:


MarijnS95:
> and pretty-print bit enum operands

Otherwise it debug-prints the same enum variant double, right?

msiglreith:
I would output `FunctionControl(INLINE | CONST)` for example in the test

rspirv/sr/storage.rs
```diff
@@ -16,10 +16,7 @@ pub struct Token<T> {
 
 impl<T> Clone for Token<T> {
     fn clone(&self) -> Self {
-        Token {
-            index: self.index,
-            marker: self.marker,
-        }
+        *self
```
MarijnS95, line 19: Does this now call `Copy`... which calls `Clone`?

MarijnS95:
Hmm for `Display` having just the variant name, but for `Debug` it seems nice to also include the type name at leaset once?

rspirv/sr/storage.rs
msiglreith: `Copy` doesn't call `Clone` it's a memcpy internally, there is an example of this pattern in the rustdocs as well
https://doc.rust-lang.org/std/marker/trait.Copy.html#how-can-i-implement-copy

msiglreith:
Switched back to  the default `Debug` impl and use `Display` instead to hide the duplicate type names, needed to touch some parts of the autogenerator

rspirv/dr/constructs.rs
```diff
@@ -392,9 +392,11 @@ mod tests {
         assert_eq!(
             format!(
                 "{}",
-                dr::Operand::FunctionControl(spirv::FunctionControl::INLINE)
+                dr::Operand::FunctionControl(
+                    spirv::FunctionControl::INLINE | spirv::FunctionControl::CONST
+                )
             ),
-            "INLINE",
+            "INLINE | CONST",
```
MarijnS95, line 399: Here for example, isn't it useful to know that we're printing an _`Operand` of variant `FunctionControl`_?

msiglreith: Ok, I adjust the testcase to add the struct name and using the default `Debug` impl. (Initially I thought this testcase was for pretty-printing in disassembly but this seems to be separate)

MarijnS95: Lovely, thanks so much!

MarijnS95 (approved):

PR #251, opened 2024-03-16
Regenerate on `sdk-1.4.309.0`

MarijnS95:


spirv/autogen_spirv.rs
```diff
@@ -2990,6 +3073,60 @@ impl core::str::FromStr for StoreCacheControl {
...
+#[allow(non_upper_case_globals)]
+impl FPEncoding {}
+impl core::str::FromStr for FPEncoding {
+    type Err = ();
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            _ => Err(()),
+        }
+    }
+}
```
MarijnS95, line 3129: This is a new optional enum operand for [`OpTypeFloat`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpTypeFloat) that's not yet defined to have any values/variants:

https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#FP_Encoding

I made some changes/hacks to `autogen` to support this variant-less enum, but need more to remove these empty `match` statements or decide if it should be an opaque enum (newtype struct on `u32`?) altogether depending on how enum values will be defined in the future.

wingertge:
Would it be possible to get this regen updated to 1.4.309 and merged so we can have a new release? [cubecl-spirv](https://github.com/tracel-ai/cubecl/tree/main/crates/cubecl-spirv) is currently forced to work off this regen branch which makes it impossible to do a release right now.

The newest tag has a small change where numeric idents (specifically `2x2`) need to be prefixed, but it's a small 2 line change to this function.
``rust
/// Converts the given string into an `Ident`, with call-site span.
pub fn as_ident(ident: &str) -> Ident {
    let first_byte = ident.bytes().next().unwrap();
    // Escape bitfield variant that starts with a number
    if first_byte.is_ascii_digit() {
        Ident::new(&format!("_{ident}"), Span::call_site())
    } else {
        Ident::new(ident, Span::call_site())
    }
}
```

MarijnS95:
Unfortunately `1.4.309` is not tagged yet (just a branch for now), and it requires some manual changes before the crate compiles properly. Will take a little bit of effort to update that.

wingertge:
> Unfortunately `1.4.309` is not tagged yet (just a branch for now), and it requires some manual changes before the crate compiles properly. Will take a little bit of effort to update that.

Ah, I didn't check the tag i just assumed since it's a finalized branch it also had a tag. What are the changes required to make it compile? I have it set up on a branch and it seemed like the only change needed was the prefix on the `2x2` bitfield value. Is it for non-default features? I can work on it if needed.

MarijnS95:
@wingertge oh I thought you relied on `1.4.309`, hence the request?

Looking at the changes in the `.json`, a maintainer fluent in `rspirv-autogen` is probably more efficient and thorough in tackling all the necessary changes.

wingertge:
@MarijnS95 1.4.309 would be nice for the `Rust` source language but it's not that big a deal, it's more about some other enums we're currently using (I would have to look up which ones). I figured I'd look at it and it seems like the main change is that aliases are now defined as a separate field instead of being duplicated entries, so the operand kind needs a slight change but it's not that huge. I'm getting it compiled now but haven't done the manual enum matches in the non-generated code.

MarijnS95:
@wingertge would you be interested in testing & reviewing this PR, if you're actively using / depending on it? That's currently the most significant blocker before we can do a merge + release.

wingertge:
@MarijnS95 I ran the full `burn` and `cubecl` tests suites on the SPIR-V backend with this new version and there's no regressions. So at least the base ops, subgroup ops, float atomics and cooperative matrix extension seem to be working perfectly. We don't really use any graphics releated ops like texture sampling so I can't test those unfortunately.

wingertge (approved):

PR #254, opened 2024-12-25
spirv: Allow crate to remain `no_std` when `(de)serialize` features are enabled

Dinnerbone:
Despite `spirv` crate being `no_std`, it pulls in serde's `std` feature by accident. This fixes that, so it can be used safely in no_std crates

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,7 +48,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all --all-targets -- -Dwarnings
+          args: --workspace --all-features --all-targets -- -Dwarnings
   test:
     name: Cargo check and test
     strategy:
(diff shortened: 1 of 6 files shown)


MarijnS95:
I'm curious how this got to be (but `git blame` isn't readily available on mobile). Most likely the crate used to be `no_std` and `serde` integration was added at some later time?

In any case I don't think we should land this in isolation, but rather spend some time adding a `no_std` check to CI as well as the relevant `clippy` lints that guide contributors to use `core` primitives over `std`.

MarijnS95:
Looks like it was added in #125, but the crate is behind a non-default feature meaning that `spirv` is already `no_std` compliant unless the `serialize` or `deserialize` features are turned on.

To that effect, I think you should rewrite your pull request title and description to explain that you have tested and confirmed that _iff_ the `(de)serialize` features are used, this crate could still be completely `no_std`-compatible while also providing `serde` support.

MarijnS95:
Actually the crate wasn't `no_std` when #125 was merged yet. This only happened in #177 which added `#![no_std]` but disregarded the optional features and optional `serde` dependency.

I thought this should have been caught in CI if the crate was built/linted/tested using `--all-features`, but apparently it doesn't.

PR #255, opened 2025-02-14
Remove errant underscore in AABB abbreviation

mmoult:
AABB stands for Axis-Aligned Bounding Box. In one of the ray tracing flag enumerations, it is pluralized as AABBs. This confused the shouty snake case generation.

There was already a special handling for NaN, add another case for AABBs and refactor the handling into its own function in utils. This compares to what has been done prior for converting strings into Ident.

Update the autogen files to use the new naming rules.

Format all code, which made small edits to code I didn't otherwise modify.


MarijnS95 (commented):
Nice, thanks for commonizing it too!

It is probably unfeasible to `match` on a few input strings and capitalizing them by hand, right?

mmoult:
Depends on how you define unfeasible I guess. However, I think the replace after conversion approach (which I extended with another case) is more robust. Here are three potential strategies I see (in ascending order of my preference):

1) Keep a list of strings which `heck` cannot handle and their expected results. (I think this is what you mean by "capitalizing them by hand", but correct me if I am wrong). This would *work*, but it would require us to update the list more frequently than the other approaches and may include duplicate info (only the 'NaN' part needs to be specially handled, but all other parts would be enumerated too, for example).
2) Replace known abbreviations then feed into `heck`. For example, 'NaN' -> 'Nan' and 'AABBs' -> 'Aabbs'. This should work too, but requires more thought to give `heck` an input we know it will accept (and therefore, we are more dependent on `heck`'s behavior). Also, requires a replacement at runtime for instances of all known abbreviations, even if `heck` changes in the future to support them.
3) [Current] Feed the input into `heck` then replace known errors. I think post-processing is more predictable because there is no black box between our handling and the output. Also, the inverse situation described in point 2: if `heck` changes to handle `AABBs`, for example, then extra replacements we have will be unused. Assuming `heck` does the replacement more efficiently than our checking code, this would be ideal.

If you feel strongly about using another approach (including any I didn't enumerate), I could change the patch to do that, but I think it is best as-is.

mmoult:
What is next? I can be patient if people need more time to review- I just want to make sure that it isn't waiting on me.

cwfitzgerald:
Hey! Sorry about the extreme delay, I've now taken over primary maintainership here. If you can rebase on the latest we can get this in!

mmoult:
Ok. Working on the rebase now. Should be done soon.

mmoult:
Should be ready to go now. Please approve and merge.

MarijnS95:
> I've now taken over primary maintainership here.

@cwfitzgerald how should this be read? I've repeatedly kept this repository up to date while spending lots of blood/sweat/tears fending off poor external contributions or otherwise fixing them to be able to perform releases, most recently a few months ago (and a necessary regenerate is already lined up).

cwfitzgerald:
> how should this be read?

@MarijnS95 Poorly. I did not think through the implications of what I was saying here. I apologize for trivializing and under-appreciating the work you do keeping this library and others up to snuff and overstepping onto a situation I didn't fully understand. I definitely understand how thankless it can feel to be on the maintenance side of a library, and should not be perpetuating that. This reflects poorly on both myself and the wgpu maintainers as a whole, and will be sure to be mindful of the implications of my words going forward.

My goal (and the goal of the wgpu maintainers in general, having interest in this repository) is to take some of the maintenance burden off you so it doesn't need to be your sole responsibility to unblock wgpu; our needs shouldn't require you scramble to get a release out. By depending so directly on (r)spirv we should also help participate in keeping it going. I don't think this venture has been successful and in fact has almost certainly made maintaining rspirv _more_ work for you rather than less as evidenced at least by #278.

How can I/we help you in a way that isn't going to add extra burden on you? If you'd rather we leave well enough alone, that's alright too.

For when things inevitably do come up and we need to coordinate, what is the best way to contact you?

MarijnS95:
> How can I/we help you in a way that isn't going to add extra burden on you? If you'd rather we leave well enough alone, that's alright too.

@cwfitzgerald general maintenance is definitely appreciated. I keep doing the binding updates because we rely on newer SPIR-V extensions (just like `SPIRV-Tools` - everything breaks when a too-old `SPIRV-Headers` is used), but this repo has been underappreciated for far too long and all its code and hand-written hacks get more fragile by the minute (every time an update lands).

I've somewhat tried to mitigate this by for example replacing hand-written incomplete logic with spec-based getters, but it's an uphill battle. #279 also calls for yet another system based on the spec rather than hand-coded (revert-reason for #278), but is once again complex to thread through.

Careful and thorough review of external contributions is definitely appreciated too. In the age of AI, its overly verbose descriptions easily gaslight maintainers into the completely wrong change/fix, but as evident it was just as easy with a missing description altogether :wink:. Don't base it on the merit of being open/ignored for a significant amount of time (if anything I'd close such PRs instead).

> For when things inevitably do come up and we need to coordinate, what is the best way to contact you?

GitHub pings might be most convenient, or e-mail? I used to be part of some wgpu Matrix chats but even the (supposed…) development channels were too spammy to keep up with.

cwfitzgerald:
Thanks! I appreciate the walkthrough of the current state of the project.

> In the age of AI, its overly verbose descriptions easily gaslight maintainers into the completely wrong change/fix, but as evident it was just as easy with a missing description altogether 😉

100%, we're trying to figure this out in wgpu land too, it's been "fun".

> GitHub pings might be most convenient, or e-mail?

Alright, github should be good for most things; could you send me your email (my email is public on my github), I want searching for it but couldn't find anything.

PR #258, opened 2025-04-22
[spirv]: Remove unused features from `serde`

bushrat011899:
The `serialize` and `deserialize` features will break `no_std` compatibility due to default features. Disabling those defaults allows the features within `no_std`. This will improve [wgpu/6826](https://github.com/gfx-rs/wgpu/issues/6826).

### DIFF
--- a/spirv/Cargo.toml
+++ b/spirv/Cargo.toml
@@ -20,4 +20,4 @@ path = "lib.rs"
 
 [dependencies]
 bitflags = "2.0"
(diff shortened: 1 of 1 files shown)


MarijnS95:
This looks like a duplicate of https://github.com/gfx-rs/rspirv/pull/254, which was recently improved and is more elaborate. Would you mind taking a look there? Thanks for the reminder either way :+1:

bushrat011899:
Ah brilliant, didn't see that, thanks!

PR #262, opened 2025-06-20
Regenerate on `sdk-1.4.313.0`

MarijnS95:


kvark:
LGTM, but who is generally reviewing rspirv these days?

MarijnS95:
Looking at the diff myself, I've yet to add some new `OpType`s to the `grammar` so that we don't end up with an error like #266 later, will do so soon.

MarijnS95:
It always surprised me that we had to manually add these (and have a lot of bugs like #266 as a result), even though they can be very trivially generated: #269

MarijnS95:
No this branch was already purely ahead of `master` with zero conflicts. I do always _rebase merge_ though.

cwfitzgerald:
Sorry yeah rebase merge was what I meant.

PR #268, opened 2025-09-23
DRAFT: Support lifting `OpFunctionCall` to Structured Representation

MarijnS95:
Haven't decided whether this should go to a new `ops::Op::FunctionCall` or a separate structure. `lift_op()` currently doesn't support `OpFunctionCall`, but it feels like adding it to `ops::Op` should be paired with supporting it in `lift_op()`.

### DIFF
--- a/rspirv/lift/mod.rs
+++ b/rspirv/lift/mod.rs
@@ -154,6 +154,11 @@ impl LiftContext {
                                 };
                             }
                         }
+                        spirv::Op::FunctionCall => {
+                            let op = context.lift_function_call(inst)?;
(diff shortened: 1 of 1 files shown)


MarijnS95:
@dreveman you have pinged me on two other occasions and also sent an email, and are expecting a <24h lead time before spamming completely unrelated PRs. Don't you think that's being impatient? This doesn't seem all that important, and I'm not confident that your contribution fixes what I have previously been blocked on to publish a maintained implementation to those reserved crate names.

Please keep the pings and discussions limited to the Perfetto repo.

PR #273, opened 2026-03-13
Generate and implement extended instruction handling for *all* `extinst` sets

MarijnS95:
Previous PRs made a mess out of extended instructions by adding _some_ boilerplate for them but never updating it in `ExtInstSetTracker` making it impossible to be used in the binary parser. This generalizes `Instruction` and `ExtendedInstruction` a little bit to handle both cases.

Disclaimer: most commits were written by me, but I've asked claude to help me with the indefinite churn in the last two commits.


MarijnS95:
> In hindsight I'm unhappy that the extended instructions are generalized to a `spirv::Word` to save on generic complexity down the line, while the _operands_ were added to the core `enum OperandKind` enum. That's something I'd like to flatten out.

Let claude fix that now, though in a slightly different way.

MarijnS95:
@cwfitzgerald can you review this once more?

cwfitzgerald (approved):
Just finished reviewing when you pushed - everything looks good and I understood the motivation behind the code very well, thank you for all the commentary on the commits. Please merge at will!

Also woooo more autogen!

PR #275, opened 2026-05-31
binary: Fix parse_words on big-endian targets

nadzyah:
parse_words reinterpreted the host-endian in-memory bytes of the input &[u32] as a byte buffer, but the decoder always reads words as little-endian (from_le_bytes). On big-endian hosts (e.g. s390x) this byte-swaps every word, so the SPIR-V magic number is read swapped and parsing fails with EndiannessUnsupported. This is already caught by the existing test_parse_words test, which fails on big-endian.

Keep the zero-copy reinterpretation on little-endian targets and re-serialise each word into little-endian byte order on big-endian targets via to_le_bytes.

Originally found on s390x arch when testing the `rust-rspirv` Debian package:
https://ci.debian.net/packages/r/rust-rspirv/testing/s390x/71641368/

### DIFF
--- a/rspirv/binary/parser.rs
+++ b/rspirv/binary/parser.rs
@@ -7,7 +7,7 @@ use super::{
     tracker::{Type, TypeTracker},
     DecodeError,
 };
-use std::{error, fmt, result, slice};
+use std::{error, fmt, result};
 
 use crate::grammar::OperandKind as GOpKind;
 use crate::grammar::OperandQuantifier as GOpCount;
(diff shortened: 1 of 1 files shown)


MarijnS95:
Looking at this in more detail, the assembling side doesn't seem to be affected because we only ever return `Vec<u32>` (anyone writing that to a file has to consider the host->LE conversion themselves).

On the parsing side we seem to be converting to `u8` only to convert back to `spirv::Word` (`u32`) with `from_le_bytes()` again. So this problem really only seems to exist because we provide a flawed `load_words()`/`parse_words()` `u32` helper.
I wonder if that's even useful, or if we should instead rewrite `Parser::new()` to never round-trip via `u8` because it's always reading whole `u32` anyway (and leave the optional conversion for the `u8` helper path instead...).

As long as we provide a `u8` helper loader interprets the source data as LE (and a `u32` loader that treats the source data as native-endian), I think we should also add a helper that outputs `u8` data in LE format.

PR #276, opened 2026-06-16
spirv: improve documentation detailing SPIR-V specification version

ickk:
I was a little bit confused when trying to understand what version of the spir-v specification the `spirv` crates corresponds to. 

This patch includes the SPIR-V specification version in the `spirv` crate documentation explicitly, and further clarifies that the version-metadata corresponds to the git tag of the SPIRV-HEADERS repository.


MarijnS95:
I'm taking a different approach to this in https://github.com/gfx-rs/rspirv/pull/280.

